PR #1, opened 2023-02-16
Add workspace elements

nick-mobilecoin:

PR #2, opened 2023-02-16
Generic `And` and `Or` verification steps

nick-mobilecoin:


nick-mobilecoin:
* main
    * **PR #2** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/2" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/2?utm_source=stack-comment).

awygle (approved):
This looks good.

verifier/src/lib.rs
```diff
@@ -3,10 +3,222 @@
 #![doc = include_str!("../README.md")]
 #![deny(missing_docs, missing_debug_implementations, unsafe_code)]
 
+use std::fmt::{Debug, Formatter};
+
+type Result<T> = std::result::Result<T, VerificationError>;
+
+#[derive(Debug, Eq, PartialEq)]
+/// Failed to verify: {0}.
+pub struct VerificationError(String);
```
jcape, line 12: In general I dislike the `Error(String)` pattern, since I can't do anything but print it out (e.g. what if the `TcbInfo` I've been given is simply out-of-date and I need to fetch a new one). This also means we can never communicate securely between a hardware element that doesn't support alloc and an enclave.

My preference is to have this be an associated type on `Verifier` (which is what I did in `mc-attest-verifier`)

verifier/src/lib.rs
```diff
@@ -3,10 +3,222 @@
...
+    /// When verification fails the [`VerificationError`] should contain a
+    /// message communicating the cause of the failure.
+    fn verify(&self) -> Result<()>;
+}
+
+impl Debug for dyn VerificationStep {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("VerificationStep").finish()
+    }
+}
```
jcape, line 34:
```rust
trait Verifier: Debug { /* ... */ }
```

verifier/src/lib.rs
```diff
@@ -3,10 +3,222 @@
...
+    /// The right side of this logical and instance.
+    pub fn right(&self) -> &R {
+        &self.right
+    }
+}
+
+impl<L: VerificationStep, R: VerificationStep> VerificationStep for And<L, R> {
+    fn verify(&self) -> Result<()> {
+        self.left.verify()?;
+        self.right.verify()
```
jcape, line 72: `And` should evaluate both sides and return the union of their errors (i.e. the log output should be able to say _everything_ that's wrong with the evidence).

verifier/src/lib.rs
```diff
@@ -3,10 +3,222 @@
...
+    pub fn right(&self) -> &R {
+        &self.right
+    }
+}
+
+impl<L: VerificationStep, R: VerificationStep> VerificationStep for Or<L, R> {
+    fn verify(&self) -> Result<()> {
+        self.left.verify().or_else(|_| self.right.verify())
+    }
+}
```
jcape, line 113: Similar story as `And`, we should be able to build a log message that can say "neither <left> nor <right> was true"

jcape (changes requested):

verifier/src/lib.rs
nick-mobilecoin: > My preference is to have this be an associated type on Verifier (which is what I did in mc-attest-verifier)

Where is that at, I did a quick scrape and only saw it on some hex encode/decode stuff

verifier/src/lib.rs
```diff
@@ -3,10 +3,222 @@
...
+
+impl<S: Into<String>> From<S> for VerificationError {
+    fn from(message: S) -> Self {
+        Self(message.into())
+    }
+}
+
+/// A verification step. These can chained together using the [`Or`] and [`And`]
+/// types.
+pub trait VerificationStep {
```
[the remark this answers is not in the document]
nick-mobilecoin: renamed

verifier/src/lib.rs
```diff
@@ -3,10 +3,222 @@
 #![doc = include_str!("../README.md")]
 #![deny(missing_docs, missing_debug_implementations, unsafe_code)]
 
+use std::fmt::{Debug, Formatter};
```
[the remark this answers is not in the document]
nick-mobilecoin: made it no std

verifier/src/lib.rs
nick-mobilecoin: done, Thanks!

verifier/src/lib.rs
nick-mobilecoin: I don't think I'm fully following this. From the comment above:
> This also means we can never communicate securely between a hardware element that doesn't support alloc and an enclave.

It seems like we need some kind of growing structure to be able to hold all of the errors as they percolate up, or we need to say we can hold a fixed size of errors.  But either way we need some way to hold indicators that "TCBInfo out of date" and "Debug flag was on".  I can understand if we don't want the informative messages/info of "SVN version was 3, needs to be at least 5", or maybe we can choose some errors to contain smaller concrete values.

verifier/src/lib.rs
jcape: Sorry, this was in my initial stab at improving error logging: https://github.com/jcape/mobilecoin/blob/92d200058ee597fbd94dac2d002e2439a95b2636/attest/verifier/src/report_body.rs#L102

verifier/src/lib.rs
nick-mobilecoin: using CtOption with the self::error logic

verifier/src/lib.rs
nick-mobilecoin: With newer logic we should be able to walk the tree and provide display/debug logic.
However the change here seems big enough that I think we follow on with the debug/display stuff

verifier/src/lib.rs
nick-mobilecoin: using associated error type now.
Currently there is a generic VerificationError. My thought is as we pull pieces over from the Epid logic we can expand the enum that it will become.

nick-mobilecoin:
@jcape my next go at the `And` and `Or` logic is up now with https://github.com/mobilecoinfoundation/attestation/pull/2/commits/1507cc14399a4144a97875e4c9a43bdf3f2593e6

verifier/src/lib.rs
```diff
@@ -2,11 +2,334 @@
...
+    }
+}
+
+impl<L: Verifier, R: Verifier> Verifier for Or<L, R> {
+    type Error = OrError<L::Error, R::Error>;
+    fn verify(&self) -> CtOption<Self::Error> {
+        CtOption::new(
+            OrError::new(self.left.verify(), self.right.verify()),
+            1.into(),
+        )
```
jcape, line 205: Same archetype as `And<L, R>`:
```suggestion
        let left_err = self.left.verify();
        let right_err = self.right.verify();
        let both_err = left_err.is_some() & right_err.is_some();
        CtOption::new(OrError::new(left_err, right_err), both_err)
```

jcape (changes requested):
Great job, but our princess is in another castle (still a bit of work around handling `CtOption`)

verifier/src/lib.rs
```diff
@@ -2,11 +2,334 @@
...
+    }
+}
+
+impl<L: Verifier, R: Verifier> Verifier for And<L, R> {
+    type Error = AndError<L::Error, R::Error>;
+    fn verify(&self) -> CtOption<Self::Error> {
+        CtOption::new(
+            AndError::new(self.left.verify(), self.right.verify()),
+            1.into(),
+        )
```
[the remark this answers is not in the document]
nick-mobilecoin: done, used `is_some` instead of `either_err` to match `CtOption::new()`

verifier/src/lib.rs
nick-mobilecoin: done, used `is_some` instead of `either_err` to match `CtOption::new()`

verifier/src/lib.rs
```diff
@@ -2,11 +2,334 @@
...
+/// This is *not* constant time.
+impl<T: IntoResult<(), E>, E> IntoResult<(), E> for CtOption<T> {
+    fn into_result(self) -> Result<(), E> {
+        let option: Option<T> = self.into();
+        match option {
+            None => Ok(()),
+            Some(e) => e.into_result(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
nick-mobilecoin: For some reason I was thinking we needed to walk down the error tree to determine the error state, and also I didn't implement correctly to bubble up the error state.
This has been removed now

verifier/src/lib.rs
```diff
@@ -2,11 +2,260 @@
...
+    _left: CtOption<L>,
+    _right: CtOption<R>,
+}
+
+impl<L, R> AndError<L, R> {
+    /// Create a new instance
+    pub fn new(left: CtOption<L>, right: CtOption<R>) -> Self {
+        Self {
+            _left: left,
+            _right: right,
```
jcape, line 39: Is this prefixed because "written but never read"?

jcape (approved):

verifier/src/lib.rs
nick-mobilecoin: That is correct, clippy barked about it.
I think once we implement the display, this should go away

nick-mobilecoin:
@awygle bumped you for a review since this had changed significantly since when you reviewed it

PR #3, opened 2023-02-17
Port `Attributes` verification from MobileCoin repo

nick-mobilecoin:


nick-mobilecoin:
* main
      * **PR #3** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/3" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/3?utm_source=stack-comment).

verifier/src/report_body.rs
```diff
@@ -0,0 +1,177 @@
...
+}
+
+impl<'a> ReportBodyVerifier<'a> {
+    /// Create a new [`ReportBodyVerifier`] instance.
+    ///
+    /// # Arguments:
+    /// * `expected_report_body` - The expected report body.
+    /// * `report_body` - The report body to verify conforms to the
+    ///   `expected_report_body`.
+    pub fn new(expected_report_body: ReportBody, report_body: &'a ReportBody) -> Self {
```
nick-mobilecoin, line 23: I think we're still narrowing in on what the final API should be, for now this is using `ReportBody`, but we're moving toward a higher level `Evidence` or similar. I would like to flush out a few more of the lower level elements in the ReportBody and see how that flushes out.
The main reason being that there are multiple reports in the evidence and I'm not quite convinced  we want a `QEReportAttributesVerifier` and a `AppReportAttributesVerifier`

verifier/src/report_body.rs
```diff
@@ -0,0 +1,177 @@
+// Copyright (c) 2018-2023 The MobileCoin Foundation
+
+//! Verifiers which operate on the [`ReportBody`](::mc_attest_core::ReportBody)
+//! data structure.
+
+use crate::{Result, VerificationStep};
```
nick-mobilecoin, line 6: I had named the common trait `VerificationStep`.  Thinking about it, the "step" adjective feels like it's what the higher level container will see these as, not what they see themselves as. Also looking at the current Epid logic that uses `Verifier` I'm thinking this should  get renamed to `Verifier`.

Naming is hard and I'm up for other guidance

verifier/src/report_body.rs
```diff
@@ -0,0 +1,177 @@
...
+        report_body_src.attributes.flags = 0;
+        let report_body_src: ReportBody = report_body_src.into();
+        let verifier = ReportBodyVerifier::new(report_body_src.clone(), &report_body);
+        assert_eq!(
+            verifier.verify(),
+            Err(VerificationError(format!(
+                "Attributes mismatch: expected {:?}, got {:?}",
+                report_body_src.attributes(),
+                report_body.attributes()
+            )))
```
nick-mobilecoin, line 150: I think this exposes a current weakness in my desire to have a generic `AttributesVerifier` that doesn't care which report it's part of. This error message still causes a client to dig around and find where the incorrect attributes are at.
I'm thinking either I need to get on board with the `QeReportAttributesVerifier` and `AppReportAttributesVerifier` or the `ReportBody` needs to capture child errors and further decorate. `"Quoting enclave, ReportBody, Attributes mismatch: expected ...."`

verifier/src/report_body.rs
jcape: There's two steps:

1. Extract data from the Evidence.
1. Check that evidence against expected data.

So my off-the-cuff thoughts is to add a new `Accessor` trait:

```rust
trait Accessor<T> {
    fn get(evidence: &Evidence) -> T;
}

struct QeReportAttributesAccessor;
impl Accessor<Attributes> for QeReportAttributesAccessor {
    fn get(evidence: &Evidence) -> Attributes {
        /* ... */
    }
}

struct AttributesVerifier<T: Accessor<Attributes>> {
    expected: Attributes,
    _marker: PhantomData<T>
}

impl<T: Accessor<Attributes>> Verifier for Attributes<T> {
    type Error = AttributesError;
    fn verify(&self, evidence: &Evidence) -> Result<(), AttributesError> {
        if (self.expected == T::get(evidence)) {
            Ok(())
        } else {
            Err(AttributesError)
        }
    }

type QeReportAttributesVerifier = AttributesVerifier<QeReportAttributesAccessor>;
type AppAttributesVerifier = AttributesVerifier<AppAttributesAccessor>;
```

verifier/src/report_body.rs
nick-mobilecoin: This was fixed in #2

verifier/src/report_body.rs
nick-mobilecoin: Captured improvements in #8

verifier/src/report_body.rs
nick-mobilecoin: Added as a task to #7, unless there is strong preference for now, I would like to get one or 2 more sub pieces in and then see how it looks.

verifier/src/report_body.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+}
+
+impl Verifier for AttributesVerifier<'_> {
+    type Error = VerificationError;
+    fn verify(&self) -> CtOption<Self::Error> {
+        let expected = self.expected_attributes;
+        let actual = self.report_body.attributes();
+        // TODO - This should be a constant time comparison.
+        let is_some = if expected == actual { 0 } else { 1 };
```
nick-mobilecoin, line 67: I'm up for suggestions here, it may be that in order to implement this I need to flush out how the verifiers are actually getting passed the data they're verifying.
it may be that the verifier is a newtype wrapper around the `Attributes` and then it can implement `ConstantTimeEq`

jcape: I'd say we should just `impl ConstantTimeEq for Attributes` in `core-types`, honestly.

jcape (approved):
This is fine, but my strong opinion is that we will need to supply the full Evidence to everything if we want to have this built off a string / use `And`+`Or`.

varsha888 (approved):

verifier/src/report_body.rs
nick-mobilecoin: created https://github.com/mobilecoinfoundation/sgx/issues/280

PR #10, opened 2023-02-23
Add no alloc build to CI

nick-mobilecoin:

PR #9, opened 2023-02-23
Port in ConfigId verifier for ReportBody

nick-mobilecoin:

PR #4, opened 2023-02-17
Remove specific CODEOWNERS for metadata files

nick-mobilecoin:

PR #11, opened 2023-02-23
Rework `Verifier`s to be passed in the evidence

nick-mobilecoin:
This also changed the `ReportBodyVerifier` to use a builder pattern for
the fields to verify.

### DIFF
--- a/verifier/src/lib.rs
+++ b/verifier/src/lib.rs
@@ -34,15 +34,31 @@ pub enum VerificationError {
 
 /// A verifier. These can chained together using the [`Or`] and [`And`]
 /// types.
-pub trait Verifier: Debug {
+pub trait Verifier<T>: Debug {
     /// The error that this verification will return in failure cases.
     type Error;
 
-    /// Performs a verification operation for the [`Verifier`].
+    /// Performs a verification operation on `evidence`.
     ///
     /// In order to accommodate constant time operations this returns a
     /// [`CtOption`] instead of a [`Result`].
-    fn verify(&self) -> CtOption<Self::Error>;
+    fn verify(&self, evidence: &T) -> CtOption<Self::Error>;
+
+    /// Or this verifier with another.
+    fn or<U, V: Verifier<U>>(self, other: V) -> Or<Self, V>
+    where
+        Self: Sized,
+    {
+        Or::new(self, other)
+    }
+
+    /// And this verifier with another.
+    fn and<U, V: Verifier<U>>(self, other: V) -> And<Self, V>
+    where
+        Self: Sized,
+    {
+        And::new(self, other)
+    }
 }
 
 /// An error that occurs during an `and` operation.
@@ -95,11 +111,11 @@ impl<L, R> And<L, R> {
     }
 }
 
-impl<L: Verifier, R: Verifier> Verifier for And<L, R> {
+impl<T, L: Verifier<T>, R: Verifier<T>> Verifier<T> for And<L, R> {
     type Error = AndError<L::Error, R::Error>;
-    fn verify(&self) -> CtOption<Self::Error> {
-        let left_err = self.left.verify();
-        let right_err = self.right.verify();
+    fn verify(&self, evidence: &T) -> CtOption<Self::Error> {
+        let left_err = self.left.verify(evidence);
+        let right_err = self.right.verify(evidence);
         let is_some = left_err.is_some() | right_err.is_some();
         CtOption::new(AndError::new(left_err, right_err), is_some)
     }
@@ -155,34 +171,34 @@ impl<L, R> Or<L, R> {
     }
 }
 
-impl<L: Verifier, R: Verifier> Verifier for Or<L, R> {
+impl<T, L: Verifier<T>, R: Verifier<T>> Verifier<T> for Or<L, R> {
     type Error = OrError<L::Error, R::Error>;
-    fn verify(&self) -> CtOption<Self::Error> {
-        let left_err = self.left.verify();
-        let right_err = self.right.verify();
+    fn verify(&self, evidence: &T) -> CtOption<Self::Error> {
+        let left_err = self.left.verify(evidence);
+        let right_err = self.right.verify(evidence);
         let is_some = left_err.is_some() & right_err.is_some();
         CtOption::new(OrError::new(left_err, right_err), is_some)
     }
 }
 
 /// Will always succeed for the [`Verifier::verify()`] operation.
-#[derive(Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, Hash, PartialEq, Default)]
 pub struct AlwaysTrue;
 
-impl Verifier for AlwaysTrue {
+impl<T> Verifier<T> for AlwaysTrue {
     type Error = VerificationError;
-    fn verify(&self) -> CtOption<Self::Error> {
+    fn verify(&self, _evidence: &T) -> CtOption<Self::Error> {
         CtOption::new(VerificationError::General, 0.into())
     }
 }
 
 /// Will always fail for the [`Verifier::verify()`] operation.
-#[derive(Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, Hash, PartialEq, Default)]
 pub struct AlwaysFalse;
 
-impl Verifier for AlwaysFalse {
+impl<T> Verifier<T> for AlwaysFalse {
     type Error = VerificationError;
-    fn verify(&self) -> CtOption<Self::Error> {
+    fn verify(&self, _evidence: &T) -> CtOption<Self::Error> {
         CtOption::new(VerificationError::General, 1.into())
     }
 }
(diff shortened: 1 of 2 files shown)


nick-mobilecoin:
* main
      * **PR #3** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/3" target="_blank">Graphite</a> 
        * **PR #9** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/9" target="_blank">Graphite</a> 
          * **PR #10** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/10" target="_blank">Graphite</a> 
            * **PR #11** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/11" target="_blank">Graphite</a>  👈
    * **PR #4** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/4" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/11?utm_source=stack-comment).

verifier/src/lib.rs
```diff
@@ -34,15 +34,31 @@ pub enum VerificationError {
...
-    /// Performs a verification operation for the [`Verifier`].
+    /// Performs a verification operation on `evidence`.
     ///
     /// In order to accommodate constant time operations this returns a
     /// [`CtOption`] instead of a [`Result`].
-    fn verify(&self) -> CtOption<Self::Error>;
+    fn verify(&self, evidence: &T) -> CtOption<Self::Error>;
+
+    /// Or this verifier with another.
+    fn or<U, V: Verifier<U>>(self, other: V) -> Or<Self, V>
```
nick-mobilecoin, line 48: Up for thoughts here I thought it would be nicer to `verifier.and(other)` instead of having to `And::new(verifier, other)` all the time.

verifier/src/report_body.rs
```diff
@@ -2,72 +2,95 @@
...
+}
+
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
-pub struct ReportBodyVerifier<'a> {
-    expected_report_body: ReportBody,
-    report_body: &'a ReportBody,
+pub enum ReportBodyVerifierKind {
+    AttributesVerifier(AttributesVerifier),
+    ConfigIdVerifier(ConfigIdVerifier),
+    AlwaysTrue(AlwaysTrue),
```
nick-mobilecoin, line 20: As I was working on when/where the _evidence_ should be passed I realized that a client may only want to verify portions of `ReportBody`. Since the verification tree is a concrete type via generics it has to be known at compile time and thus can't be built up dynamically. In order to accomodate this I brought in the kind enum from the main mobilecoin repo.

I'm up for a more rusty alternative, keeping in mind that one of the desires was to have this be no_alloc.

verifier/src/report_body.rs
```diff
@@ -2,72 +2,95 @@
...
-            AttributesVerifier::new(self.expected_report_body.attributes(), self.report_body);
-        let config_id_verifier =
-            ConfigIdVerifier::new(self.expected_report_body.config_id(), self.report_body);
-        let and_verifier = And::new(attributes_verifier, config_id_verifier);
-        and_verifier.verify()
+    fn verify(&self, evidence: &ReportBody) -> CtOption<Self::Error> {
+        let verifier = self
+            .attributes_verifier
+            .clone()
+            .and(self.config_id_verifier.clone());
```
nick-mobilecoin, line 64: Not fond of this cloning, it may be that I need to use refernces and lifetimes here.

verifier/src/report_body.rs
```diff
@@ -2,72 +2,95 @@
...
-            expected_report_body,
-            report_body,
+impl Verifier<ReportBody> for ReportBodyVerifierKind {
+    type Error = VerificationError;
+
+    fn verify(&self, evidence: &ReportBody) -> CtOption<Self::Error> {
+        match self {
+            Self::AttributesVerifier(verifier) => verifier.verify(evidence),
+            Self::ConfigIdVerifier(verifier) => verifier.verify(evidence),
+            Self::AlwaysTrue(verifier) => verifier.verify(evidence),
```
nick-mobilecoin, line 30: An off the cuff suggestion in https://github.com/mobilecoinfoundation/attestation/pull/3#discussion_r1110210689 was to have an accessor trait, but I'm thinking that mixes multiple abstraction levels.
My thought is the higher level logic that says to "Verify the app report body" will know where the app report body is and can pass it to the "ReportBodyVerifier" and it will know where the QE report body is and pass it to another "ReportBodyVerifier" instance. I'm thinking `QuoteVerifier`'s `verify()` will be provided a Quote and then it will pass the individual report bodies to the appropriate verifiers.

Having said what I said above, now that I'm looking at this I'm wondering if this should be:
```suggestion
            Self::AttributesVerifier(verifier) => verifier.verify(evidence.attributes()),
            Self::ConfigIdVerifier(verifier) => verifier.verify(evidence.config_id()),
            Self::AlwaysTrue(verifier) => verifier.verify(evidence),
```

verifier/src/lib.rs
nick-mobilecoin: Thinking some more this might be harder to understand when one builds the tree, i.e.
```rust
a.and(b).and(c)
```
This will result in the following tree, which may not be as intuitive.
```
           And
           /\
          /  \
         /    \
       And     C
       /\
      /  \
     /    \
    A      B
```

verifier/src/report_body.rs
nick-mobilecoin: I've made the change in #12. 
I started adding in another member of ReportBody and it started to feel clunky having to initialize a ReportBody to use the verifier.

verifier/src/lib.rs
```diff
@@ -219,14 +239,14 @@ mod tests {
     #[test]
     fn and_succeeds() {
         let and = And::new(AlwaysTrue, AlwaysTrue);
-        let verification = and.verify();
+        let verification = and.verify(NO_EVIDENCE);
         assert_eq!(verification.is_none().unwrap_u8(), 1);
```
NotGyro, line 243: I realize this might be out of the scope of the PR and referring back to prior work, but why does the return type of `.verify()` use Option-like semantics rather than something more like a `Result<(), E>`? It's probable that there's a very good reason for this (for example, constant-time), but I figured I might as well double-check.

verifier/src/report_body.rs
NotGyro: Doing things like this with an enum is plenty Rust-idiomatic. It might become tricky to work with if we then need to have the type system know something about this at compile-time later, but that seems unlikely given the purpose of this enum is so the actual concrete type can vary at run-time.

verifier/src/report_body.rs
NotGyro: Could `.and()` be refactored to take `&self` rather than `self`? I realize that would also change the signature of `And<R, L>` so it might be outside of the scope of this PR.

NotGyro (changes requested):
This mostly looks good to me - changes might not be needed, but there were some comments for your consideration.

verifier/src/report_body.rs
nick-mobilecoin: This has now been removed, in favor of the the `Accessor` traite

verifier/src/report_body.rs
nick-mobilecoin: Possibly, I'm a little reluctant as the idea is to build up a tree of verifiers and if they're references then one will need to keep all of the tree nodes around instead of passing the tree around and forgetting about it.

verifier/src/report_body.rs
nick-mobilecoin: Moved this pr to use an Accessor trait.

verifier/src/lib.rs
nick-mobilecoin: > It's probable that there's a very good reason for this (for example, constant-time)
Correct it's for constant time.  Subtle doesn't have a Result type, probably because one would need to create the OK and the Error type to be constant time. So this is using CtOption<Error> as the result, where None (or 0) == success.
Up for ideas on how best to communicate this.
There is:
```rust
    /// In order to accommodate constant time operations this returns a
    /// [`CtOption`] instead of a [`Result`].
    fn verify(&self, evidence: &T) -> CtOption<Self::Error>;
```
On the `Verifier` trait, but perhaps it needs to be expanded more.

NotGyro (approved):
This looks good to me

verifier/src/lib.rs
jcape: Back in the day, GCC and Clang would turn `a && b && c` into either `a && (b && c)` or `(a && b) && c` based on RNG by default, so I don't think this is a huge deal ;-)

jcape (approved):
This looks good, but I thought we'd decided we were going to have Verifier always take `AttestationEvidence`?

verifier/src/lib.rs
nick-mobilecoin: Wow, I thought the C spec specified it should be (a && b) && c

nick-mobilecoin:
> This looks good, but I thought we'd decided we were going to have Verifier always take `AttestationEvidence`?

We had originally, however since I got it working with generics I figured it would be better to keep it generic unless we run into issues.
this makes it easier to test, and often easier to test == easier to use.

Issue #7, opened 2023-02-13
Move report verification stuff into new repo


jcape:
- [x] #3 
- [x] #9 
- [x] #10 
- [x] #11
- [x] #18
- [x] #20
- [x] #21
- [x] #22
- [x] #34
  - [x] https://github.com/mobilecoinfoundation/sgx/issues/286
  - [x] https://github.com/mobilecoinfoundation/sgx/issues/285
  - [x] https://github.com/mobilecoinfoundation/sgx/issues/283
- [x] #42 
- [x] #38 
- [x] #39 
- [x] #40 
- [x] #41 
- [x] #30

PR #23, opened 2023-02-28
Add `Verifier` for `MrEnclave`

nick-mobilecoin:


verifier/src/report_body.rs
```diff
@@ -219,13 +231,57 @@ impl<T: Accessor<MiscellaneousSelect>> Verifier<T> for MiscellaneousSelectVerifi
...
+}
+
+impl Accessor<MrEnclave> for ReportBody {
+    fn get(&self) -> MrEnclave {
+        let Measurement::MrEnclave(mr_enclave) = self.mr_enclave() else {
+            panic!("`mr_enclave()` should return a Measurement::MrEnclave");
+        };
+        mr_enclave
+    }
+}
```
nick-mobilecoin, line 259: This is here until, https://github.com/mobilecoinfoundation/sgx/issues/286.
Added as a task to #7

verifier/src/report_body.rs
```diff
@@ -24,6 +27,17 @@ pub trait Accessor<T>: Debug {
     fn get(&self) -> T;
 }
 
+/// [`Accessor`] for returning Self, i.e. T -> T
```
awygle, line 30: What's the point of this macro? It's only used in one place right now, will it be used more in other places? What's the difference compared to just calling `.clone()`?

awygle (commented):
This looks fine, but I don't really understand the point of that macro.

verifier/src/report_body.rs
nick-mobilecoin: The main reason for:
```rust
impl Accessor<T> for T {
    fn get(&self) -> T {
        self.clone()
     }
}
```
(and now that I think of it, I realize I can implement a generic impl instead of a macro... will do on a subsequent PR)

Anyway,
In order to be able to pass an instance of `Thing` to the `ThingVerifier` there needs to be an `Accessor` trait for `Thing`.
```
impl<T: Accessor<Thing>> Verifier<T> for ThingVerifier {

    fn verify(&self, evidence: &T) -> ...
```

awygle (approved):

jcape (approved):

nick-mobilecoin:
bumping @awygle due to restack

awygle (approved):

PR #30, opened 2023-03-03
Consolidate equality logic behind Generic interface

nick-mobilecoin:
Previously the equality operation of verification was implemented
manually for each verifier type. Now the equality verifications all use
the same generic implementation.

PR #26, opened 2023-02-28
Use generic impl instead of `self_accessor` macro

nick-mobilecoin:

PR #25, opened 2023-02-28
Add `Verifier` for `ReportData`

nick-mobilecoin:

PR #24, opened 2023-02-28
Add `Verifier` for `MrSigner`

nick-mobilecoin:

PR #31, opened 2023-03-04
Add display capability to `Verifier` output

nick-mobilecoin:
Ignore the branch name of "debug" this is a "Display" impl

### DIFF
--- a/verifier/src/lib.rs
+++ b/verifier/src/lib.rs
@@ -86,17 +86,44 @@ pub enum VerificationError {
 
 impl DisplayableError for VerificationError {}
 
-#[derive(Debug, Clone)]
-/// A [`CtOption`] wrapped to implement the [`Display`] trait.
-pub struct DisplayableCtOption<T>(CtOption<T>);
+/// Trait to convert a [`CtOption<T>`] into a [`CtOptionDisplay<'a, T>`].
+///
+/// # Examples
+/// ```
+/// use subtle::CtOption;
+/// use mc_attestation_verifier::{VerificationError, DisplayableCtOption};
+///
+/// let ct_option = CtOption::new(VerificationError::General, 0.into());
+///
+/// assert_eq!(format!("{}", ct_option.display()), "Passed");
+/// ```
+pub trait DisplayableCtOption<'a, T> {
+    /// Returns an object that implements [`Display`] for a wrapped
+    /// [`CtOption<T>`].
+    #[must_use = "this does not display the [`CtOption<T>`], \
+                  it returns an object that can be displayed"]
+    fn display(&'a self) -> CtOptionDisplay<'a, T>
+    where
+        Self: Sized;
+}
 
-impl<T> From<CtOption<T>> for DisplayableCtOption<T> {
-    fn from(ct_option: CtOption<T>) -> Self {
+impl<'a, T> DisplayableCtOption<'a, T> for CtOption<T> {
+    fn display(&'a self) -> CtOptionDisplay<'a, T> {
+        self.into()
+    }
+}
+
+#[derive(Debug, Clone)]
+/// Helper struct for displaying [`CtOption`] with [`format!`] and `{}`.
+pub struct CtOptionDisplay<'a, T>(&'a CtOption<T>);
+
+impl<'a, T> From<&'a CtOption<T>> for CtOptionDisplay<'a, T> {
+    fn from(ct_option: &'a CtOption<T>) -> Self {
         Self(ct_option)
     }
 }
 
-impl<T: DisplayableError> Display for DisplayableCtOption<T> {
+impl<'a, T: DisplayableError> Display for CtOptionDisplay<'a, T> {
     fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
         let option: Option<T> = self.0.clone().into();
         match option {
@@ -104,7 +131,7 @@ impl<T: DisplayableError> Display for DisplayableCtOption<T> {
                 true => write!(f, "{value:#}")?,
                 false => write!(f, "{value}")?,
             },
-            None => write!(f, "passed")?,
+            None => write!(f, "Passed")?,
         }
 
         Ok(())
@@ -159,14 +186,8 @@ impl<L: DisplayableError, R: DisplayableError> DisplayableError for AndError<L, 
 impl<L: DisplayableError, R: DisplayableError> Display for AndError<L, R> {
     fn fmt(&self, f: &mut Formatter) -> core::fmt::Result {
         f.debug_struct("AndError")
-            .field(
-                "left",
-                &format_args!("{:#}", DisplayableCtOption(self.left.clone())),
-            )
-            .field(
-                "right",
-                &format_args!("{:#}", DisplayableCtOption(self.right.clone())),
-            )
+            .field("left", &format_args!("{:#}", self.left.display()))
+            .field("right", &format_args!("{:#}", self.right.display()))
             .finish()
     }
 }
@@ -233,14 +254,8 @@ impl<L: DisplayableError, R: DisplayableError> DisplayableError for OrError<L, R
 impl<L: DisplayableError, R: DisplayableError> Display for OrError<L, R> {
     fn fmt(&self, f: &mut Formatter) -> core::fmt::Result {
         f.debug_struct("OrError")
-            .field(
-                "left",
-                &format_args!("{:#}", DisplayableCtOption(self.left.clone())),
-            )
-            .field(
-                "right",
-                &format_args!("{:#}", DisplayableCtOption(self.right.clone())),
-            )
+            .field("left", &format_args!("{:#}", self.left.display()))
+            .field("right", &format_args!("{:#}", self.right.display()))
             .finish()
     }
 }
@@ -418,8 +433,8 @@ mod tests {
             },
             0.into(),
         );
-        let displayable: DisplayableCtOption<_> = success.into();
-        assert_eq!(format!("{displayable}"), "passed");
+        let displayable = success.display();
+        assert_eq!(format!("{displayable}"), "Passed");
     }
 
     #[test]
@@ -431,7 +446,7 @@ mod tests {
             },
             1.into(),
         );
-        let displayable: DisplayableCtOption<_> = failure.into();
+        let displayable = failure.display();
         assert_eq!(
             format!("{displayable:}"),
             "The ISV svn value of IsvSvn(2) is less than the expected value of IsvSvn(3)"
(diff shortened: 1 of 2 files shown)


nick-mobilecoin:
* main
      * **PR #3** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/3" target="_blank">Graphite</a> 
        * **PR #9** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/9" target="_blank">Graphite</a> 
          * **PR #10** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/10" target="_blank">Graphite</a> 
            * **PR #11** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/11" target="_blank">Graphite</a> 
              * **PR #12** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/12" target="_blank">Graphite</a> 
              * **PR #18** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/18" target="_blank">Graphite</a> 
                * **PR #20** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/20" target="_blank">Graphite</a> 
                  * **PR #21** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/21" target="_blank">Graphite</a> 
                    * **PR #22** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/22" target="_blank">Graphite</a> 
                      * **PR #23** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/23" target="_blank">Graphite</a> 
                        * **PR #24** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/24" target="_blank">Graphite</a> 
                          * **PR #25** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/25" target="_blank">Graphite</a> 
                            * **PR #26** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/26" target="_blank">Graphite</a> 
                              * **PR #30** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/30" target="_blank">Graphite</a> 
                                * **PR #31** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/31" target="_blank">Graphite</a>  👈
    * **PR #4** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/4" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/31?utm_source=stack-comment).

verifier/src/lib.rs
```diff
@@ -10,67 +10,72 @@ pub use report_body::{
...
+
 /// Failed to verify.
-#[derive(Debug, Eq, PartialEq)]
+#[derive(displaydoc::Display, Debug, Eq, PartialEq, Clone)]
 pub enum VerificationError {
     /// A general error.
     General,
-    /// The attributes did not match expected:{expected} actual:{actual}
+    /// Forced failure via `AlwaysFalse`
+    AlwaysFalse,
```
nick-mobilecoin, line 28: Added this as it felt odd to have "A general error" in the tree output instead of a more informative always false indicator, still probably a bit difficult for a user to understand though :(

verifier/src/lib.rs
```diff
@@ -10,67 +10,72 @@ pub use report_body::{
...
+    /// The attributes did not match expected:{expected:?} actual:{actual:?}
```
nick-mobilecoin, line 29: Had to use `Debug` for these, created https://github.com/mobilecoinfoundation/sgx/issues/290

verifier/src/lib.rs
```diff
@@ -79,11 +84,38 @@ pub enum VerificationError {
     },
 }
 
+impl DisplayableError for VerificationError {}
+
+#[derive(Debug, Clone)]
+/// A [`CtOption`] wrapped to implement the [`Display`] trait.
+pub struct DisplayableCtOption<T>(CtOption<T>);
```
nick-mobilecoin, line 91: Not sure if there is a better way to do this. feels a bit clunky to wrap it up to display, perhaps the Verifier should return a `DisplayableCtOption`, or maybe a free function that can print/display `CtOption` would be better :shrug:

verifier/src/lib.rs
```diff
@@ -79,11 +84,38 @@ pub enum VerificationError {
...
+    }
+}
+
+impl<T: DisplayableError> Display for DisplayableCtOption<T> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        let option: Option<T> = self.0.clone().into();
+        match option {
+            Some(value) => match f.alternate() {
+                true => write!(f, "{value:#}")?,
+                false => write!(f, "{value}")?,
```
nick-mobilecoin, line 105: Initially I had
```suggestion
                true => write!(f, "Failure: {value:#}")?,
```
but the `Failure: ` string looked noisy in the output and I don't think "passed" is likely to get confused for an error.

verifier/src/lib.rs
```diff
@@ -79,11 +84,38 @@ pub enum VerificationError {
...
+            },
+            None => write!(f, "passed")?,
```
nick-mobilecoin, line 107: Chose to not dig into passing options, figured `Debug` could be used for that, even though one would have to look at 0 vs 1 for the ct option.

verifier/src/lib.rs
```diff
@@ -10,67 +10,72 @@ pub use report_body::{
...
 
-use core::fmt::Debug;
+use core::fmt::{Debug, Display, Formatter};
 use mc_sgx_core_types::{
     Attributes, ConfigId, ConfigSvn, IsvSvn, MiscellaneousSelect, MrEnclave, MrSigner, ReportData,
 };
 use subtle::CtOption;
 
+/// An error that implements the [`Display`] trait.
+pub trait DisplayableError: Display + Clone {}
```
nick-mobilecoin, line 20: Needed to have `Clone` here as `CtOption` consumes itself anytime one wants to get the underlying value.

verifier/src/lib.rs
jcape: I'd say return the `CtOption`, but implement `CtOptionDisplay<'_>`, ala [std::path::Display](https://doc.rust-lang.org/std/path/struct.Display.html). From there, you can monkey-patch a `display()` method via a local trait, providing a similar interface to [`Path::display()`](https://doc.rust-lang.org/std/path/struct.Path.html#method.display).

nick-mobilecoin: Updated to use a trait with `display()` method.

verifier/src/lib.rs
```diff
@@ -169,19 +236,27 @@ impl<T, L: Verifier<T>, R: Verifier<T>> Verifier<T> for And<L, R> {
...
+}
+
+impl<L: DisplayableError, R: DisplayableError> DisplayableError for OrError<L, R> {}
+
+impl<L: DisplayableError, R: DisplayableError> Display for OrError<L, R> {
+    fn fmt(&self, f: &mut Formatter) -> core::fmt::Result {
+        f.debug_struct("OrError")
+            .field("left", &format_args!("{:#}", self.left.display()))
+            .field("right", &format_args!("{:#}", self.right.display()))
+            .finish()
```
jcape, line 259: Please don't use `debug_struct` for `Display`.

verifier/src/lib.rs
```diff
@@ -109,19 +168,27 @@ pub trait Verifier<T>: Debug {
...
+}
+
+impl<L: DisplayableError, R: DisplayableError> DisplayableError for AndError<L, R> {}
+
+impl<L: DisplayableError, R: DisplayableError> Display for AndError<L, R> {
+    fn fmt(&self, f: &mut Formatter) -> core::fmt::Result {
+        f.debug_struct("AndError")
+            .field("left", &format_args!("{:#}", self.left.display()))
+            .field("right", &format_args!("{:#}", self.right.display()))
+            .finish()
```
jcape, line 191: See comments on `OrError`

verifier/src/lib.rs
jcape, line 259: Please don't use `debug_struct` for `Display`. Recommend:

```
Neither of these conditions were satisfied:
  - {left:#}
  - {right:#}
```

jcape (changes requested):
LGTM, one thing that might be interesting to add to the Displayable is checkmarks.

varsha888 (approved):

verifier/src/lib.rs
nick-mobilecoin: Updated, but don't think it's quite what was being suggested

verifier/src/lib.rs
nick-mobilecoin: updated but don't think it's quite what was desired

verifier/src/lib.rs
```diff
@@ -346,4 +494,128 @@ mod tests {
...
+        let expected = r#"
+            OrError:
+              - [ ]
+                AndError:
+                  - [x]
+                    Passed
+                  - [ ]
+                    The ISV svn value of IsvSvn(1) is less than the expected value of IsvSvn(3)
+              - [ ]
+                The MiscellaneousSelect did not match expected:MiscellaneousSelect(2) actual:MiscellaneousSelect(3)"#;
```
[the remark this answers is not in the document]
nick-mobilecoin: I'm not sure I understand what `CtOption` is getting us if we want to keep the underlying verifier around?
`CtOption` will only provide the wrapped `value` if it `is_some()` this means in the non failing cases we can't get to the value which is the reason for the current "Passed" text. So it seems like all we're leveraging is the `Choice` part of the `CtOption`.

I think if we move to a dedicated type we can be more ergonomic and avoid the `CtOption` limitations.
```rust
pub struct VerificationResult {
    is_ok: Choice,
    error: VerificationError,
    verifier: &'a Verifier
}
```

verifier/src/lib.rs
```diff
@@ -109,19 +184,81 @@ pub trait Verifier<T>: Debug {
...
+}
+
+/// Turn a `Choice` into a markdown checkbox
+///
+/// Will return one of:
+/// ```raw
+///     - [ ]
+///     - [x]
+/// ```
+fn choice_to_checkbox(choice: Choice) -> &'static str {
```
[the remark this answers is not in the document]
nick-mobilecoin:
```console
error[E0015]: cannot call non-const fn `<bool as From<Choice>>::from` in constant functions
   --> verifier/src/lib.rs:208:8
    |
208 |     if bool::from(choice) {
    |        ^^^^^^^^^^^^^^^^^^
```

It looks like none of the `Choice` methods are const

PR #48, opened 2023-03-19
Add `VerificationTreeDisplay`

nick-mobilecoin:
The `VerificationTreeDisplay` allows one to display the output of a
verification, showing which phases succeeded, which failed and what the
values were for each phase.

### DIFF
--- a/verifier/src/lib.rs
+++ b/verifier/src/lib.rs
@@ -19,14 +19,24 @@ pub use report_body::{
 
 use crate::struct_name::SpacedStructName;
 use core::fmt::{Debug, Display, Formatter};
-use mc_sgx_core_types::{
-    Attributes, ConfigId, ConfigSvn, CpuSvn, ExtendedProductId, FamilyId, IsvProductId, IsvSvn,
-    MiscellaneousSelect, MrEnclave, MrSigner, ReportData,
-};
 use subtle::Choice;
 
-/// Number of spaces to indent nested error messages.
-const ERROR_INDENT: usize = 2;
+/// Number of spaces to indent nested messages.
+const MESSAGE_INDENT: usize = 2;
+
+/// Success checkbox indicator
+const SUCCESS_MESSAGE_INDICATOR: &str = "- [x]";
+
+/// Failure checkbox indicator
+const FAILURE_MESSAGE_INDICATOR: &str = "- [ ]";
+
+fn choice_to_status_message(choice: Choice) -> &'static str {
+    if choice.into() {
+        SUCCESS_MESSAGE_INDICATOR
+    } else {
+        FAILURE_MESSAGE_INDICATOR
+    }
+}
 
 /// The output of a verification operation, [`Verifier::verify()`].
 #[derive(Debug, Clone)]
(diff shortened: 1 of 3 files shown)


nick-mobilecoin:
* main
  * **PR #48** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/48" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/48?utm_source=stack-comment).

verifier/src/lib.rs
```diff
@@ -13,192 +13,126 @@ pub use report_body::{
...
+}
+
+impl<V: Display, O: SpacedStructName + Debug> VerificationMessage<O> for V {
+    fn fmt_padded(
+        &self,
+        f: &mut Formatter<'_>,
+        pad: usize,
+        output: &VerificationOutput<O>,
+    ) -> core::fmt::Result {
+        let is_success = output.is_success();
```
nick-mobilecoin, line 113: This might be locking in an implementation too much. I had 3 copies of this so genericised it.
It may be that we want this to key off of a marker trait

nick-mobilecoin:
> I thought we'd discussed that the primary source of the string at each print should be the display of the verifier...?

I think that's the self in this , https://github.com/mobilecoinfoundation/attestation/pull/48/files#diff-fe2f7ab2bdcef7981ced2b0119aa6146527d7beebe5bf3808e8652f6663e1eadR115
```rust
write!(f, "{:pad$}{status} {self}", "")?;
```

There are some places where that isn't done:
- [`And`](https://github.com/mobilecoinfoundation/attestation/pull/48/files#diff-fe2f7ab2bdcef7981ced2b0119aa6146527d7beebe5bf3808e8652f6663e1eadR222) It felt a little odd to have "the following" if `And` was displayed by itself, perhaps the message "Both of the following must be true" should be changed to "Both must be true".
- [`Or`](https://github.com/mobilecoinfoundation/attestation/pull/48/files#diff-fe2f7ab2bdcef7981ced2b0119aa6146527d7beebe5bf3808e8652f6663e1eadR299) similar to above
- `Not`, `AlwaysTrue`, `AlwaysFalse`, I'll work on changing these
- [`AttributesVerifier`](https://github.com/mobilecoinfoundation/attestation/pull/48/files#diff-e06ecba8c109a9d2d77b167638e3b55dc9a02dc8a1b646cd0ab2bf25ec100becR182) similar to `And`, the separate line by line comparison doesn't seem to lend itself to the Verifier display well
- [`MrSignerVerifier`](https://github.com/mobilecoinfoundation/attestation/pull/48/files#diff-e06ecba8c109a9d2d77b167638e3b55dc9a02dc8a1b646cd0ab2bf25ec100becR365) similar to the `And`/`AttributesVerifier`

nick-mobilecoin:
my bad @cbeck88, I should have moved this to a draft, I broke it up and refactored in #86 and #87, with a follow on soon

nick-mobilecoin:
requested a re-review, @cbeck88, since this got refactored some with the new main line as well as #86 and #87

cbeck88:
Np

PR #49, opened 2023-03-22
Add logic to decode and verify certificate

nick-mobilecoin:
A DER encoded certificate can now be decoded and have it's signature
verified


nick-mobilecoin:
* main
  * **PR #49** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/49" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/49?utm_source=stack-comment).

.github/workflows/ci.yaml
```diff
@@ -155,37 +155,6 @@ jobs:
         with:
           files: lcov.info
```
nick-mobilecoin, line 157: x509-cert is `no_std` but it uses alloc, https://github.com/RustCrypto/formats/blob/master/x509-cert/src/lib.rs#L18

I don't recall the full impetus for wanting no alloc in the verify crate.  This may get back to the mention of having a trait for getting access to the certs and keep an alloc crate behind a feature flag

verifier/src/lib.rs
```diff
@@ -13,6 +14,8 @@ pub use report_body::{
     MiscellaneousSelectVerifier, MrEnclaveVerifier, MrSignerVerifier, ReportDataVerifier,
 };
 
+pub use certs::{Error as CertificateError, UnverifiedCertificate, VerifiedCertificate};
```
nick-mobilecoin, line 17: These are exported here to make cargo happy. Current htinking is that these will be hid behind
`CertificationData::verify` or similar

verifier/src/certs.rs
```diff
@@ -0,0 +1,314 @@
...
+}
+
+impl<'a> UnverifiedCertificate<'a> {
+    /// Verify the certificate signature.
+    pub fn verify(self, key: &VerifyingKey) -> Result<VerifiedCertificate<'a>> {
+        let tbs_length = self.certificate.tbs_certificate.encoded_len()?;
+        let tbs_size = u32::from(tbs_length) as usize;
+        let tbs_contents = &self.der_bytes[TBS_OFFSET..tbs_size + TBS_OFFSET];
+        key.verify(tbs_contents, &self.signature)
+            .map_err(|_| Error::SignatureVerification)?;
```
nick-mobilecoin, line 65: This returns a result and we'll need to chain multiple certificate verifications.  I'm not sure this will be able to be constant time.

Thinking ahead:
I'm also curious what we show the user in failure cases once we get to verifying the actual quote signature. Guessing we want to show "key" and then in failure cases "hash, expected signature".
But how will the certificates fit into the verifiers tree?  
```console
- [ ] QE report signature:
  - [ ] Cert 1 key (and time)
  - [ ] Cert 2 key (and time)
  - [ ] Cert 3 key (and time)
  - [ ] QE signing key
```

.github/workflows/ci.yaml
jcape: The intent was to allow us to communicate between a hardware wallet and a secure enclave. but one thing is that cert verification can just live behind a provider, and we only provide one when "alloc" is enabled.

verifier/src/certs.rs
jcape: In this case, simply saying "Certificate chain verified successfully" and "Certificate chain was signed by a trusted CA" is probably good enough.

.github/workflows/ci.yaml
nick-mobilecoin: Updated to have an `alloc` feature

NotGyro (approved):
Looks good to me.

verifier/src/certs.rs
```diff
@@ -0,0 +1,314 @@
...
+        BgsqhkiG+E0BDQECCAIBADAQBgsqhkiG+E0BDQECCQIBADAQBgsqhkiG+E0BDQEC
+        CgIBADAQBgsqhkiG+E0BDQECCwIBADAQBgsqhkiG+E0BDQECDAIBADAQBgsqhkiG
+        +E0BDQECDQIBADAQBgsqhkiG+E0BDQECDgIBADAQBgsqhkiG+E0BDQECDwIBADAQ
+        BgsqhkiG+E0BDQECEAIBADAQBgsqhkiG+E0BDQECEQIBCzAfBgsqhkiG+E0BDQEC
+        EgQQERECBAGABgAAAAAAAAAAADAQBgoqhkiG+E0BDQEDBAIAADAUBgoqhkiG+E0B
+        DQEEBAYAkG7VAAAwDwYKKoZIhvhNAQ0BBQoBADAKBggqhkjOPQQDAgNJADBGAiEA
+        1XJi0ht4hw8YtC6E4rYscp9bF+7UOhVGeKePA5TW2FQCIQCIUAaewOuWOIvstZN4
+        V8Zu8NFCC4vFg+cZqO6QfezEaA==
+        -----END CERTIFICATE-----
+        ";
```
samdealy, line 141: Consider putting this const and the following consts in a separate file.

verifier/src/certs.rs
```diff
@@ -0,0 +1,314 @@
...
+        Ur9QGzknBqwwDgYDVR0PAQH/BAQDAgEGMBIGA1UdEwEB/wQIMAYBAf8CAQEwCgYI
+        KoZIzj0EAwIDSQAwRgIhAOW/5QkR+S9CiSDcNoowLuPRLsWGf/Yi7GSX94BgwTwg
+        AiEA4J0lrHoMs+Xo5o/sX6O9QWxHRAvZUGOdRQ7cvqRXaqI=
+        -----END CERTIFICATE-----
+        ";
+
+    #[parameterized(
+        root = { ROOT_CA },
+        intermediate = { INTERMEDIATE_CA },
+        leaf = { LEAF_CERT },
```
samdealy, line 184: Consider removing parameterization given you only have one value for each parameter.

samdealy (approved):

verifier/src/certs.rs
awygle: Agree with Sam - this is a prime case for `include_str!`.

awygle (approved):

verifier/src/certs.rs
nick-mobilecoin: Moved PEM's out to separate files and used include_str!

verifier/src/certs.rs
nick-mobilecoin: I may be misunderstanding the comment. The main usecase for parameterization is the same tests steps with different inputs, in this case the different inputs are the different PEMs.

Are you thinking there is a better alternative when there is only one argument?

samdealy (approved):

nick-mobilecoin:
@nick-mobilecoin queued this pull request to merge with [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/49).

nick-mobilecoin:
@nick-mobilecoin merged this pull request with [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/49).

PR #51, opened 2023-03-23
Add verification of certificate time

nick-mobilecoin:
Previously the certificate time was not checked to be within range of
the `Validity`, https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.5.
Now the certificate time is verified to be within the `Validity`


nick-mobilecoin:
* main
  * **PR #49** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/49" target="_blank">Graphite</a> 
    * **PR #51** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/51" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/51?utm_source=stack-comment).

verifier/src/certs.rs
```diff
@@ -56,20 +59,50 @@ pub struct VerifiedCertificate<'a> {
...
+    /// Verify the certificate signature and time are valid.
+    ///
+    /// # Arguments
+    /// - `key` - The public key to verify the certificate signature with
+    /// - `unix_time` - The duration since UNIX_EPOCH. This value can have
+    ///   `nanos`.
+    pub fn verify(
+        self,
+        key: &VerifyingKey,
+        unix_time: Duration,
```
awygle, line 71: This looks like it should take an `Instant` instead of a `Duration` but I'm pretty sure that's just Rust time API weirdness...

awygle (approved):
This looks fine, modulo the Rust time API is a bit weird. One enhancement could be a "Duration but I promise it's relative to the Unix epoch" type, but... eh. Probably not worth it.

verifier/src/certs.rs
```diff
@@ -56,20 +59,50 @@ pub struct VerifiedCertificate<'a> {
...
+        let not_after = validity.not_after.to_unix_duration();
+
+        // Per https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.5 time is
+        // inclusive
+        if (not_before..=not_after).contains(&unix_time) {
+            Ok(())
+        } else {
+            Err(Error::CertificateExpired)
+        }
+    }
```
samdealy, line 105: Could we do something like:
```suggestion
        if not_before <= unix_time && unix_time <= not_after {
            Ok(())
        } else {
            Err(Error::CertificateExpired)
        }
    }
```
Creating a `Range` object to perform an equivalency check seems a little wonky to me.

samdealy (approved):

verifier/src/certs.rs
nick-mobilecoin: In order to support `Instant` we would need to bring some kind of time capability into the enclave, however there isn't a trusted way to do this.  Instead we have the time passed so we don't bring in another untrusted interface into the enclave.

verifier/src/certs.rs
nick-mobilecoin: Changed logic to have a clearer error for before time.

verifier/src/certs.rs
nick-mobilecoin: Updated the docs on unix_time to try and help clarify the usage

samdealy (approved):

nick-mobilecoin:
@nick-mobilecoin queued this pull request to merge with [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/51).

nick-mobilecoin:
[Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/51) rebased this pull request as part of a merge.

nick-mobilecoin:
@samdealy Can you review again? Sorry looks like graphite web app rebased, pretty sure I didn't check the box (I need to disable the require re-review check)

nick-mobilecoin:
[Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/51) couldn't merge this PR because it was not satisfying all requirements.

PR #54, opened 2023-03-31
Implement `Display` traits for Verifiers

samdealy:

### DIFF
--- a/verifier/src/lib.rs
+++ b/verifier/src/lib.rs
@@ -4,8 +4,6 @@
 #![deny(missing_docs, missing_debug_implementations, unsafe_code)]
 #![no_std]
 
-extern crate alloc;
-
 mod report_body;
 mod struct_name;
 
--- a/verifier/src/lib.rs
+++ b/verifier/src/lib.rs
@@ -4,6 +4,8 @@
 #![deny(missing_docs, missing_debug_implementations, unsafe_code)]
 #![no_std]
 
+extern crate alloc;
+
 mod report_body;
 mod struct_name;
 
--- a/verifier/src/report_body.rs
+++ b/verifier/src/report_body.rs
@@ -2,8 +2,9 @@
 
 //! Verifiers which operate on the [`ReportBody`]
 
+use crate::struct_name::SpacedStructName;
 use crate::{VerificationError, Verifier};
-use core::fmt::Debug;
+use core::fmt::{Debug, Display, Formatter};
 use mc_sgx_core_types::{
     Attributes, ConfigId, ConfigSvn, CpuSvn, ExtendedProductId, FamilyId, IsvProductId, IsvSvn,
     MiscellaneousSelect, MrEnclave, MrSigner, ReportBody, ReportData,
@@ -79,6 +80,15 @@ pub struct EqualityVerifier<T> {
 impl<T> EqualityVerifier<T> {
     pub fn new(expected: T) -> Self {
         Self { expected }
+    }
+}
+
+impl<T> Display for EqualityVerifier<T>
+where
+    T: SpacedStructName + Debug,
+{
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        write!(f, "The {} is {:?}", T::spaced_struct_name(), self.expected)
     }
 }
 
@@ -339,6 +349,7 @@ impl<E: Accessor<ReportData>> Verifier<E> for ReportDataVerifier {
 mod test {
     use super::*;
     use crate::And;
+    use alloc::format;
     use mc_sgx_core_sys_types::{
         sgx_attributes_t, sgx_cpu_svn_t, sgx_measurement_t, sgx_report_body_t, sgx_report_data_t,
     };
@@ -877,5 +888,19 @@ mod test {
                 .unwrap_u8(),
             1
         );
+    }
+
+    #[test]
+    fn attributes_verifier_error_display() {
+        let attributes = Attributes::from(REPORT_BODY_SRC.attributes);
+        let verifier = AttributesVerifier::new(attributes);
+        let display_string = format!("{}", verifier);
+        let expected = format!(
+            "The {} is {:?}",
+            Attributes::spaced_struct_name(),
+            attributes
+        );
+
+        assert_eq!(display_string, expected)
     }
 }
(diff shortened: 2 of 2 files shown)


verifier/src/lib.rs
```diff
@@ -4,6 +4,8 @@
 #![deny(missing_docs, missing_debug_implementations, unsafe_code)]
 #![no_std]
 
+extern crate alloc;
+
```
nick-mobilecoin, line 8: :x: This crate should support [no-std::no-alloc](https://github.com/mobilecoinfoundation/attestation/blob/935c6a18c854cba93ae373c89c426a80ca5136c9/verifier/Cargo.toml#L6)

Suggest removing here and only adding to the test that needs it. Guessing the IDE added this originally.
```suggestion
```

verifier/src/report_body.rs
```diff
@@ -339,6 +349,7 @@ impl<E: Accessor<ReportData>> Verifier<E> for ReportDataVerifier {
 mod test {
     use super::*;
```
nick-mobilecoin, line 350: :wrench: To go along with removing extern alloc from lib.rs

```suggestion
mod test {
    extern crate alloc;
    
    use super::*;
```

verifier/src/report_body.rs
```diff
@@ -878,4 +889,18 @@ mod test {
             1
         );
     }
+
+    #[test]
+    fn attributes_verifier_error_display() {
```
nick-mobilecoin, line 894: :wrench: This doesn't look to be "error" related. Maybe the drive for this change was to display the verifier in error circumstances, but this looks to be testing the display, not necessarily errors
```suggestion
    fn attributes_verifier_display() {
```

verifier/src/report_body.rs
```diff
@@ -878,4 +889,18 @@ mod test {
...
+        let attributes = Attributes::from(REPORT_BODY_SRC.attributes);
+        let verifier = AttributesVerifier::new(attributes);
+        let display_string = format!("{}", verifier);
+        let expected = format!(
+            "The {} is {:?}",
+            Attributes::spaced_struct_name(),
+            attributes
+        );
```
nick-mobilecoin, line 902: :wrench: This should probably be using the display version (once the SGX side is available)
```suggestion
        let expected = format!("The {} is {attributes}", Attributes::spaced_struct_name());
```

nick-mobilecoin (changes requested):
Crate should be no-alloc _normally_.  Another PR is adding alloc as a feature.
Looks like this isn't going to pass CI until the underlying SGX types get updated.

verifier/src/report_body.rs
samdealy: Done.

verifier/src/report_body.rs
samdealy: Done.

verifier/src/lib.rs
samdealy: Done.

samdealy:
* main
  * **PR #54** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/54" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/54?utm_source=stack-comment).

nick-mobilecoin (approved):

verifier/src/report_body.rs
```diff
@@ -82,6 +83,15 @@ impl<T> EqualityVerifier<T> {
     }
 }
 
+impl<T> Display for EqualityVerifier<T>
+where
+    T: SpacedStructName + Display,
+{
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        write!(f, "The {} is {}", T::spaced_struct_name(), self.expected)
```
nick-mobilecoin, line 91: :thought_balloon: : If I'm understanding correctly we're still going to want a tree like output for what was verified.
It seems like this saying "The X is 3" is a verification output and not a verifier display.
IFF the plan is to use a tree output then the output in failure conditions would look like
```console
- [ ] The X is 3
  - [ ] The X was 2
```
Perhaps this should be something like "The expected {} is {}" or "The {} should be {}" so that the failed output would be
```
- [ ] The X should be 3
  - [ ] The actual X was 2
```
Then a success version being
```
- [x] The X should be 3
  - [x] The actual X was 3
```

verifier/src/report_body.rs
```diff
@@ -110,6 +120,15 @@ impl<T> GreaterThanEqualVerifier<T> {
     }
 }
 
+impl<T> Display for GreaterThanEqualVerifier<T>
+where
+    T: SpacedStructName + Display,
+{
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        write!(f, "The {} is {}", T::spaced_struct_name(), self.expected)
```
nick-mobilecoin, line 128: :wrench: Similar to above this should probably be something like
```suggestion
        write!(f, "The {} should be greater than or equal to {}", T::spaced_struct_name(), self.expected)
```

verifier/src/report_body.rs
```diff
@@ -878,4 +900,117 @@ mod test {
...
+
+    #[test]
+    fn attributes_verifier_display() {
+        let inner = Attributes::from(REPORT_BODY_SRC.attributes);
+        let verifier = AttributesVerifier::new(inner);
+        let display_string = format!("{}", verifier);
+
+        let expected = format!("The {} is {inner}", Attributes::spaced_struct_name());
+
+        assert_eq!(display_string, expected)
```
nick-mobilecoin, line 912: :shrug: Not sure it's any better, it would require one to `use alloc::string::ToString`
```suggestion

        let expected = format!("The {} is {inner}", Attributes::spaced_struct_name());

        assert_eq!(verifier.to_string(), expected)
```

nick-mobilecoin (approved):
LGTM, mainly thoughts/questions about what we want the display string to be

verifier/src/report_body.rs
samdealy: Agreed, will change according to the comment below.

verifier/src/report_body.rs
samdealy: Done.

verifier/src/report_body.rs
samdealy: Done.

jcape (approved):
LGTM, my only suggestion would be to remove the `"The "` from the strings, and return a capitalized string from SpacedStructName.

PR #84, opened 2023-05-10
Add parsing of the `tcbInfo` inner field

nick-mobilecoin:
Add parsing of the `tcbInfo` field of the retrieved TCB data from
<https://api.trustedservices.intel.com/sgx/certification/v4/tcb?fmspc={}>

PR #83, opened 2023-05-09
Add TCB struct and parser

nick-mobilecoin:
Add the higher level TCB parser which separates the `tcb_info` from the
`signature` in the provided TCB data from
<https://api.trustedservices.intel.com/sgx/certification/v4/tcb?fmspc={}>

PR #85, opened 2023-05-10
Add verification of TCB data

nick-mobilecoin:
The TCB data retrieved from
<https://api.trustedservices.intel.com/sgx/certification/v4/tcb?fmspc={}>,
can now be verified to have the correct signature and time period.


nick-mobilecoin:
* main
  * **PR #83** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/83" target="_blank">Graphite</a> 
    * **PR #84** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/84" target="_blank">Graphite</a> 
      * **PR #85** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/85" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/85?utm_source=stack-comment).

verifier/Cargo.toml
```diff
@@ -14,23 +14,27 @@ repository = { workspace = true }
 rust-version = { workspace = true }
 
 [features]
-alloc = ["pem-rfc7468/alloc", "dep:const-oid", "dep:p256", "dep:x509-cert", "dep:serde_json", "dep:serde"]
+alloc = ["pem-rfc7468/alloc", "dep:const-oid", "dep:p256", "dep:x509-cert", "tcb"]
+tcb = ["dep:p256", "dep:serde_json", "dep:serde", "dep:der", "dep:hex"]
```
nick-mobilecoin, line 18: I broke this out to it's own feature for now.
The alloc one was getting big.

verifier/src/tcb.rs
```diff
@@ -31,16 +31,36 @@
...
 
 /// Error parsing TCB(Trusted Computing Base) info
 #[derive(displaydoc::Display, Debug)]
 pub enum Error {
+    /// Error converting from DER
+    Der(der::Error),
     /// Error parsing TCB(Trusted Computing Base) json info: {0}
     Serde(serde_json::Error),
+    /// Error decoding the signature in the TCB data
+    SignatureDecodeError,
```
nick-mobilecoin, line 51: Up for suggestions here as `serde_json::Error` doesn't derive `PartialEq` which means this error isn't and it's a bit ergonomic in tests, which means it may be  clunky for clients

verifier/src/tcb.rs
```diff
@@ -116,17 +137,58 @@ pub struct TcbComponent {
...
+    ///
+    /// # Arguments
+    /// - `key` - The public key to verify the `tcbInfo` signature with
+    /// - `time` - The current system time
+    ///   This is expected to be generated by:
+    ///     ```ignore
+    ///     let time = DateTime::from_system_time(SystemTime::now()).unwrap();
+    ///     ```
+    ///   or equivalent
+    pub fn verify(self, key: &VerifyingKey, time: DateTime) -> Result<(), Error> {
```
nick-mobilecoin, line 159: Up for suggestions on this API.
When doing the x509 I followed the pattern of `UnverfiedFoo.verify() -> VerifiedFoo` and then left most of the behavior on the verified version to kind of nudge clients to verify before looking at anything.  I kind of like the idea, not sure how it is in practice.  Also depends how we integerate with the verifier.

verifier/src/tcb.rs
```diff
@@ -116,17 +137,58 @@ pub struct TcbComponent {
...
-#[derive(Debug, Deserialize)]
+#[derive(Debug, Deserialize, Clone)]
 #[serde(rename_all = "camelCase")]
-pub struct TcbInfoRaw<'a> {
-    tcb_info: &'a RawValue,
-    signature: &'a str,
+pub struct TcbInfoRaw {
+    tcb_info: Box<RawValue>,
+    #[serde(with = "hex")]
+    signature: Vec<u8>,
```
nick-mobilecoin, line 145: A couple things:
1. Adding the `serde(with`, or perhaps it was just the owned Vec, I started to run into lifetime issues with the `'de` lifetime in the `Deserialize` derive macro.  I struggled a bit and finally punted to a `Box<RawValue>`
2. I updated this to serde with, so that it would fail to parse to hex, sooner rather than later. I wanted to for the issue and next update `DateTime` but only the `chrono` versions which don't support `PartialOrd` were available to derive.  We could use chrono and then compare durations if we wanted...

verifier/src/tcb.rs
```diff
@@ -336,4 +418,117 @@ mod tests {
...
+    fn fails_before_issue_date() {
+        let key = tcb_verifying_key();
+        let tcb_json = include_str!("../data/tests/fmspc_00906ED50000_2023_05_10.json");
+        let raw_tcb = TcbInfoRaw::try_from(tcb_json).expect("Failed to parse raw TCB");
+
+        let time = "2023-05-10T13:43:26Z"
+            .parse::<DateTime>()
+            .expect("Failed to parse time");
+
+        assert_eq!(raw_tcb.verify(&key, time).is_err(), true);
```
nick-mobilecoin, line 447: This goes back to the `serde_json::Error` ideally I think this should be ensuring that this is returning the not yet valid error, but without `PartialEq` it's a compiler failure.

verifier/src/tcb.rs
cbeck88: >  Adding the serde(with, or perhaps it was just the owned Vec, I started to run into lifetime issues with the 'de lifetime in the Deserialize derive macro. I struggled a bit and finally punted to a Box<RawValue>

Sometimes this can be fixed with implementing `DeserializeOwned` instead?

cbeck88: The `Box<RawValue>` seems a bit funky, I want to play with this a bit myself, maybe we can make it a little cleaner

verifier/src/tcb.rs
cbeck88: one thing we could do is use `assert_matches!` maybe instead?

cbeck88: not a blocker

verifier/src/tcb.rs
cbeck88: yeah what i've often done is given up on errors have `PartialEq` because libraries don't consistently implement it, and i tend to use `assert_matches!` or similar when I want to test this stuff.

I think actual users of this API will use a match statement if they want to case out on the errors, so it's not a big deal for them if `PartialEq` is missing I think

verifier/src/tcb.rs
nick-mobilecoin: :facepalm: Not sure what I was thinking I knew about assert matches, anyway used `assert!(matches!(` because `assert_matches!` is still nightly and wasn't doing any extra logic on the value.

verifier/src/tcb.rs
nick-mobilecoin: updated tests to use `assert!(matches!(`

verifier/src/tcb.rs
nick-mobilecoin: This should be addressed in https://github.com/mobilecoinfoundation/attestation/pull/85/commits/e60468596073a3bbf20e7536a3d9b1b91955ac2b, by using the [`#[serde(borrow)]`](https://serde.rs/lifetimes.html#borrowing-data-in-a-derived-impl) attribute.

cbeck88 (approved):

PR #88, opened 2023-05-15
Decode TCB values in PCK leaf certificate

nick-mobilecoin:
Add the logic to decode TCB data off of the PCK leaf certificate. The
TCB values specific to an enclave are stored in the PCK leaf certificate
as OID extensions. Most importantly this contains the FMSPC value which
can be used to query for the advisories associated with these TCB
values at
<https://api.trustedservices.intel.com/sgx/certification/v4/tcb?fmspc={}>

PR #87, opened 2023-05-11
Convert from `CtOption` to `VerificationOutput`

nick-mobilecoin:
Previously the `CtOption` type was used as the return value for
`Verifier::verify()`. This resulted in the inability to access the value
used during the verification. Now a `VerificationOutput` output value
which provides access to the verification value(s).

This is a preparation step to being able to display the entire verification tree including all expected and actual values.

PR #86, opened 2023-05-11
Move common types to lib.rs

nick-mobilecoin:
Move the following generic types and traits from `report_body.rs` to
`lib.rs`:

- `EqualityVerifier`
- `GreaterThanEqualVerifier`
- `Accessor`
- `IntoVerificationError`

This move allows for other modules to make use of them.

PR #89, opened 2023-05-16
Add logic to get `Advisories` from `TcbInfo`

nick-mobilecoin:
The `Advisories` corresponding to a given TCB can now be retrieved from
the `TcbInfo`.

### DIFF
--- a/verifier/src/tcb.rs
+++ b/verifier/src/tcb.rs
@@ -64,6 +64,10 @@ pub enum Error {
     TcbInfoNotYetValid,
     /// TCB info expired
     TcbInfoExpired,
+    /// Asking for TCB levels for a different FMSPC
+    FmspcMismatch,
+    /// The TCB level reported does not match an entry in the TCB info data.
+    UnsupportedTcbLevel,
 }
 
 impl From<der::Error> for Error {
@@ -161,24 +165,30 @@ impl TcbInfo {
     /// >      TCB level. Otherwise, move to the next item on TCB Levels list.
     /// > 4. If no TCB level matches your SGX PCK Certificate, your TCB Level is
     /// >    not supported.
-    fn advisories(&self, pck_tcb: &PckTcb) -> Option<Advisories> {
+    ///
+    /// # Errors
+    /// - `Error::FmspcMismatch` if the `fmspc` in `self` does not match the one
+    ///   in `pck_tcb`.
+    /// - `Error::UnsupportedTcbLevel` if the TCB level reported is not found in
+    ///   self.
+    fn advisories(&self, pck_tcb: &PckTcb) -> Result<Advisories, Error> {
         // `self` should have been retrieved via
         // <https://api.trustedservices.intel.com/sgx/certification/v4/tcb?fmspc={}>
         // and the `pck_tcb.fmspc`. Failure here should rarely happen, but we
         // still check to ensure the client didn't get mixed up.
         if self.fmspc != pck_tcb.fmspc {
-            return None;
+            return Err(Error::FmspcMismatch);
         }
 
         for level in &self.tcb_levels {
             if level.tcb.is_corresponding_level(pck_tcb) {
-                return Some(Advisories {
+                return Ok(Advisories {
                     ids: level.advisory_ids.clone(),
                     status: level.tcb_status,
                 });
             }
         }
-        None
+        Err(Error::UnsupportedTcbLevel)
     }
 }
 
@@ -744,7 +754,8 @@ mod tests {
             status,
         };
 
-        assert_eq!(tcb_info.advisories(&tcb), Some(expected_advisories));
+        let actual_advisories = tcb_info.advisories(&tcb).expect("Failed to get advisories");
+        assert_eq!(actual_advisories, expected_advisories);
     }
 
     #[parameterized(
@@ -762,7 +773,10 @@ mod tests {
             fmspc: tcb_info.fmspc.clone(),
         };
 
-        assert_eq!(tcb_info.advisories(&tcb), None);
+        assert!(matches!(
+            tcb_info.advisories(&tcb),
+            Err(Error::UnsupportedTcbLevel)
+        ));
     }
 
     #[test]
@@ -786,7 +800,10 @@ mod tests {
             fmspc,
         };
 
-        assert_eq!(tcb_info.advisories(&tcb), None);
+        assert!(matches!(
+            tcb_info.advisories(&tcb),
+            Err(Error::FmspcMismatch)
+        ));
     }
 
     #[parameterized(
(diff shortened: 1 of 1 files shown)


nick-mobilecoin:
* main
  * **PR #83** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/83" target="_blank">Graphite</a> 
    * **PR #84** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/84" target="_blank">Graphite</a> 
      * **PR #85** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/85" target="_blank">Graphite</a> 
        * **PR #86** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/86" target="_blank">Graphite</a> 
          * **PR #87** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/87" target="_blank">Graphite</a> 
            * **PR #48** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/48" target="_blank">Graphite</a> 
              * **PR #89** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/89" target="_blank">Graphite</a>  👈
              * **PR #88** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/88" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/89?utm_source=stack-comment).

verifier/src/tcb.rs
```diff
@@ -33,12 +33,23 @@ extern crate alloc;
...
 use alloc::vec::Vec;
+use core::cmp::Ordering;
 use der::DateTime;
 use p256::ecdsa::signature::Verifier;
 use p256::ecdsa::{Signature, VerifyingKey};
 use serde::Deserialize;
 use serde_json::value::RawValue;
 
+// TODO: remove this once https://github.com/mobilecoinfoundation/sgx/pull/329
+//  and https://github.com/mobilecoinfoundation/sgx/pull/330 come in.
```
nick-mobilecoin, line 44: much as this TODO says, I envision that once the type makes its way into the SGX repo then this will go away and the SGX repo types will be used.

nick-mobilecoin:
Moving to draft. this has `cmp_svns()` which even in [parital order](https://en.wikipedia.org/wiki/Partially_ordered_set#Partial_order) isn't valid becuase it requires, [Antisymmetry](https://en.wikipedia.org/wiki/Antisymmetric_relation) which is *not* the case for comparing the svn values

verifier/src/tcb.rs
```diff
@@ -123,6 +220,36 @@ impl TryFrom<&str> for Tcb {
...
+    /// >      Certificate (from 01 to 16) with the corresponding values of SVNs
+    /// >      in sgxtcbcomponents array of TCB Level. If all SGX TCB Comp SVNs
+    /// >      in the certificate are greater or equal to the corresponding
+    /// >      values in TCB Level, go to 3.b, otherwise move to the next item
+    /// >      on TCB Levels list.
+    /// >   b. Compare PCESVN value retrieved from the SGX PCK certificate with
+    /// >      the corresponding value in the TCB Level. If it is greater or
+    /// >      equal to the value in TCB Level, read status assigned to this
+    /// >      TCB level. Otherwise, move to the next item on TCB Levels list.
+    fn is_corresponding_level(&self, tcb_info: &PckTcb) -> bool {
```
nick-mobilecoin, line 242: Up for suggestions on this name.
The "level" feels off here, maybe this method should be brought up into the `TcbLevel` struct...??

verifier/src/tcb.rs
```diff
@@ -388,35 +518,81 @@ mod tests {
                       "svn": 4,
                       "category": "It's a category with a type",
                       "type": "It's a type to go with the category"
+                    },
+                    {
+                      "svn": 5
```
nick-mobilecoin, line 523: With the move from a vec to an array this grew a bit bigger than ideal.
I'm leaning toward removing this tests since the `example_tcb.json` contains these optional fields and that's probably enough to test the correct behavior.

verifier/src/tcb.rs
```diff
@@ -78,13 +131,57 @@ pub struct TcbInfo {
...
+    /// >      TCB level. Otherwise, move to the next item on TCB Levels list.
+    /// > 4. If no TCB level matches your SGX PCK Certificate, your TCB Level is
+    /// >    not supported.
+    fn advisories(&self, pck_tcb: &PckTcb) -> Option<Advisories> {
+        // `self` should have been retrieved via
+        // <https://api.trustedservices.intel.com/sgx/certification/v4/tcb?fmspc={}>
+        // and the `pck_tcb.fmspc`. Failure here should rarely happen, but we
+        // still check to ensure the client didn't get mixed up.
+        if self.fmspc != pck_tcb.fmspc {
+            return None;
```
cbeck88, line 170: i wonder if we should return an actual error? to help distinguish between, your tcb does not match the level of anything in the pck, and there is an fmspc mismatch

verifier/src/tcb.rs
cbeck88: i'd suggest maybe a name like `matches_pck_tcb` ?

cbeck88 (approved):

verifier/src/tcb.rs
cbeck88: I think it's okay, there's no need to make the tests perfect -- if we have the tests and they are passing no need to futz with it more

verifier/src/tcb.rs
nick-mobilecoin: Updated to return an error

verifier/src/tcb.rs
nick-mobilecoin: I'm still chewing on this, will try to think overnight.

I may be confounding words again but `matches` makes me think `equals`, even `corresponding` kind of feels that way.
maybe `window`...
`is_tcb_level_window`?? IDK

verifier/src/tcb.rs
nick-mobilecoin: went away in #96

verifier/src/tcb.rs
nick-mobilecoin: still undecided, punting on naming for now, it's internal to the module so renaming later shouldn't be too difficult

nick-mobilecoin:
@nick-mobilecoin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/89).

nick-mobilecoin:
@nick-mobilecoin merged this pull request with [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/89).

PR #97, opened 2023-05-24
Add `Quote3Verifier`

nick-mobilecoin:
Add `Quote3Verifier` which can be used with the verifier composition to
verify the signature of a `Quote3`.

PR #98, opened 2023-05-24
Add `TcbRawInfoVerifier`

nick-mobilecoin:
Add `TcbRawInfoVerifier` which can be composed with other verifiers to
verify the signature and date validity of the `TcbRawInfo`.


nick-mobilecoin:
* main
  * **PR #97** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/97" target="_blank">Graphite</a> 
    * **PR #98** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/98" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/98?utm_source=stack-comment).

verifier/src/lib.rs
```diff
@@ -107,7 +112,7 @@ impl<'a, V, O> VerificationTreeDisplay<'a, V, O> {
     }
 }
 
-impl<'a, V: VerificationMessage<O>, O: Clone> Display for VerificationTreeDisplay<'a, V, O> {
+impl<'a, V: VerificationMessage<O>, O> Display for VerificationTreeDisplay<'a, V, O> {
```
nick-mobilecoin, line 115: This `Clone` was a stray requirement from eariler development it's not needed and was prohibiting the usage of `serde_json::Error` since that error doesn't support clone.

verifier/src/tcb.rs
```diff
@@ -278,10 +280,79 @@ impl<'a> TryFrom<&'a str> for TcbInfoRaw<'a> {
     }
 }
 
+/// Verifier for ensuring a raw TCB was signed with the provided key
```
nick-mobilecoin, line 283: I kind of think that the `TcbInfoRawVerifier` is all that should be in this crate, and there should be a separate crate which contains the other TCB types.
Not sure though, and also not sure what to name it. Part of the reason is this file is getting a bit big and another part is that most of the other modules are verifiers only with the types being verified living elsewhere.

At the same time I think we can get by this way for now.

verifier/src/tcb.rs
```diff
@@ -278,10 +280,79 @@ impl<'a> TryFrom<&'a str> for TcbInfoRaw<'a> {
...
+impl<'a, E: Accessor<TcbInfoRaw<'a>>> Verifier<E> for TcbInfoRawVerifier {
+    type Value = Error;
+    fn verify(&self, evidence: &E) -> VerificationOutput<Self::Value> {
+        let tcb_raw = evidence.get();
+        let result = tcb_raw.verify(&self.key, self.time);
+        let is_success = result.is_ok() as u8;
+
+        // The error will only ever be looked at in failure so the initial
+        // value shouldn't matter.
+        let mut error = Error::SignatureDecodeError;
```
nick-mobilecoin, line 322: Not sure I like this initializing to an error when there may not be one. Technically there is nothing stopping someone from accessing this value when `is_success()` is true.
It could be that we add an unknown or similar, I wouldn't want an `Error::NoError` that feels odd.

eranrund: `Option<Error>`?

eranrund (approved):

verifier/src/tcb.rs
nick-mobilecoin: yup, that's better, Thanks!
I think I avoid `expect()` more than I should to help use and enforce `Option` and `Result` types

nick-mobilecoin:
@nick-mobilecoin merged this pull request with [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/98).

PR #99, opened 2023-05-26
Add wrapping container for `Quote3` and `TcbInfoRaw`

nick-mobilecoin:
Add a new wrapping container, `Evidence` that takes a `Quote3` and a
`TcbInfoRaw`. The `Evidence` can be used with the majority of the
`Verifier` implementations. This allows one to compose one verifier and
use the `Evidence` in the `verify()` method.

### Motivation

When working on an example for the verifiers it felt like there was quite a bit of work clients needed to do to get things right. The intent is to make it less error prone and easier for clients to work with.

This change may likely need some discussion about if this path is the correct way to go.

### DIFF
--- a/verifier/Cargo.toml
+++ b/verifier/Cargo.toml
@@ -14,9 +14,8 @@ repository = { workspace = true }
 rust-version = { workspace = true }
 
 [features]
-tcb = ["dep:serde_json", "dep:serde", "dep:der", "dep:hex", "advisories", "mc-sgx-dcap-types/tcb"]
-advisories = ["dep:serde_json"]
-x509 = ["dep:mbedtls", "dep:x509-cert"]
+tcb = ["dep:serde_json", "dep:serde", "dep:der", "dep:hex", "mc-sgx-dcap-types/tcb", "x509-cert/pem"]
+x509 = ["dep:mbedtls", "x509-cert"]
 
 [dependencies]
 der = { version = "0.7.5", default-features = false, optional = true }
@@ -33,6 +32,7 @@ x509-cert = { version = "0.2.0", default-features = false, optional = true }
 
 [dev-dependencies]
 mc-sgx-core-sys-types = "0.6.1"
+mc-sgx-dcap-sys-types = "0.6.1"
 textwrap = "0.16.0"
 x509-cert = { version = "0.2.0", default-features = false, features = ["pem"] }
 yare = "1.0.2"
--- a/verifier/src/error.rs
+++ b/verifier/src/error.rs
@@ -1,0 +1,50 @@
+// Copyright (c) 2023 The MobileCoin Foundation
+
+//! Errors that can occur during verification
+
+use mc_sgx_dcap_types::TcbError;
+
+pub type Result<T> = core::result::Result<T, Error>;
+
+/// Error working with quote evidence
+#[derive(displaydoc::Display, Debug)]
+pub enum Error {
+    /// Error converting from DER {0}
+    Der(der::Error),
+    /// Error parsing TCB(Trusted Computing Base) json info: {0}
+    Serde(serde_json::Error),
+    /// Error decoding the signature in the TCB data
+    SignatureDecodeError,
+    /// Error verifying the signature
+    SignatureVerification,
+    /// TCB info not yet valid
+    TcbInfoNotYetValid,
+    /// TCB info expired
+    TcbInfoExpired,
+    /// Asking for TCB levels for a different FMSPC
+    FmspcMismatch,
+    /// The TCB level reported does not match an entry in the TCB info data.
+    UnsupportedTcbLevel,
+    /// Failure to get the TCB info from a quote {0}
+    Quote3TcbInfo(TcbError),
+    /// Unsupported quote certification data, should be `PckCertificateChain`
+    UnsupportedQuoteCertificationData,
+}
+
+impl From<der::Error> for Error {
+    fn from(e: der::Error) -> Self {
+        Error::Der(e)
+    }
+}
+
+impl From<serde_json::Error> for Error {
+    fn from(e: serde_json::Error) -> Self {
+        Error::Serde(e)
+    }
+}
+
+impl From<TcbError> for Error {
+    fn from(e: TcbError) -> Self {
+        Error::Quote3TcbInfo(e)
+    }
+}
--- a/verifier/src/evidence.rs
+++ b/verifier/src/evidence.rs
@@ -1,0 +1,336 @@
+// Copyright (c) 2023 The MobileCoin Foundation
+
+//! The full set of evidence needed for attesting a quote
+
+use crate::{Accessor, Advisories, Error, Result, TcbInfo, TcbInfoRaw};
+use der::DecodePem;
+use mc_sgx_core_types::{
+    Attributes, ConfigId, ConfigSvn, CpuSvn, ExtendedProductId, FamilyId, IsvProductId, IsvSvn,
+    MiscellaneousSelect, MrEnclave, MrSigner, ReportData,
+};
+use mc_sgx_dcap_types::{CertificationData, Quote3, TcbInfo as QuoteTcbInfo};
+use x509_cert::Certificate;
+
+/// The full set of evidence needed for verifying a quote
+///
+/// A wrapping container for a `Quote3` and a `TcbInfoRaw`. This can be used
+/// with the majority of the `Verifier` implementations from this crate.
+/// This allows one to compose one verifier and use an [`Evidence`] instance in
+/// the `verify()` method.
+///
+/// Importantly this will derive the [`Advisories`] related to the provided
+/// `quote` and `tcb_info_raw`, so that one can verify the allowed advisories.
+#[derive(Debug)]
+pub struct Evidence<'a, Q> {
+    quote: Quote3<Q>,
+    tcb_info_raw: TcbInfoRaw<'a>,
+    advisories: Advisories,
+}
+
+impl<'a, Q: AsRef<[u8]>> Evidence<'a, Q> {
+    /// Create a new instance
+    pub fn new(quote: Quote3<Q>, tcb_info_raw: TcbInfoRaw<'a>) -> Result<Self> {
+        let quote_tcb_info = tcb_info_try_from_quote(&quote)?;
+        let tcb_info = TcbInfo::try_from(&tcb_info_raw)?;
+        let advisories = tcb_info.advisories(&quote_tcb_info)?;
+        Ok(Self {
+            quote,
+            tcb_info_raw,
+            advisories,
+        })
+    }
+}
+
+impl<'a, Q> Accessor<TcbInfoRaw<'a>> for Evidence<'a, Q> {
+    fn get(&self) -> TcbInfoRaw<'a> {
+        self.tcb_info_raw.clone()
+    }
+}
+
+impl<'a, Q: Clone> Accessor<Quote3<Q>> for Evidence<'a, Q> {
+    fn get(&self) -> Quote3<Q> {
+        self.quote.clone()
+    }
+}
+
+impl<'a, Q> Accessor<Advisories> for Evidence<'a, Q> {
+    fn get(&self) -> Advisories {
+        self.advisories.clone()
+    }
+}
+
+// TODO think this should go in tcb.rs of `mc-sgx-dcap-types`
+fn tcb_info_try_from_quote<Q: AsRef<[u8]>>(quote: &Quote3<Q>) -> Result<QuoteTcbInfo> {
+    let signature_data = quote.signature_data();
+    let certification_data = signature_data.certification_data();
+    let CertificationData::PckCertificateChain(pem_chain) = certification_data else {
+        return Err(Error::UnsupportedQuoteCertificationData);
+    };
+    let leaf_cert = pem_chain
+        .into_iter()
+        .next()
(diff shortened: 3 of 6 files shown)


nick-mobilecoin:
* main
  * **PR #99** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/99" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/99?utm_source=stack-comment).

verifier/Cargo.toml
```diff
@@ -14,9 +14,8 @@ repository = { workspace = true }
 rust-version = { workspace = true }
 
 [features]
-tcb = ["dep:serde_json", "dep:serde", "dep:der", "dep:hex", "advisories", "mc-sgx-dcap-types/tcb"]
-advisories = ["dep:serde_json"]
-x509 = ["dep:mbedtls", "dep:x509-cert"]
+tcb = ["dep:serde_json", "dep:serde", "dep:der", "dep:hex", "mc-sgx-dcap-types/tcb", "x509-cert/pem"]
```
nick-mobilecoin, line 17: I dropped `advisories` feature, I think it's really a part of the tcb feature, i.e. it's highly unlikely someone will want only advisories, especially since the advisories are contained in the tcb json.

verifier/src/error.rs
```diff
@@ -0,0 +1,50 @@
...
+
+use mc_sgx_dcap_types::TcbError;
+
+pub type Result<T> = core::result::Result<T, Error>;
+
+/// Error working with quote evidence
+#[derive(displaydoc::Display, Debug)]
+pub enum Error {
+    /// Error converting from DER {0}
+    Der(der::Error),
```
nick-mobilecoin, line 13: I contemplated placing the x509Error in here, but I think the x509 certificate logic is separate functionality, it just happens to live in this crate as a convenient place to host the functionality.

verifier/src/evidence.rs
```diff
@@ -0,0 +1,334 @@
...
+///
+/// A wrapping container for a `Quote3` and a `TcbInfoRaw`. This can be used
+/// with the majority of the `Verifier` implementations from this crate.
+/// This allows one to compose one verifier and use an [`Evidence`] instance in
+/// the `verify()` method.
+///
+/// Importantly this will derive the [`Advisories`] related to the provided
+/// `quote` and `tcb_info_raw`, so that one can verify the allowed advisories.
+#[derive(Debug)]
+pub struct Evidence<'a, Q> {
```
nick-mobilecoin, line 24: Up for suggestions on this name, it feels a bit too generic, but it may also be my Cisms creeping in and wanting to namespace the object name.

verifier/src/evidence.rs
```diff
@@ -0,0 +1,334 @@
...
+    tcb_info_raw: TcbInfoRaw<'a>,
+    advisories: Advisories,
+}
+
+impl<'a, Q: AsRef<[u8]>> Evidence<'a, Q> {
+    /// Create a new instance
+    pub fn new(quote: Quote3<Q>, tcb_info_raw: TcbInfoRaw<'a>) -> Result<Self> {
+        let quote_tcb_info = tcb_info_try_from_quote(&quote)?;
+        let tcb_info = TcbInfo::try_from(&tcb_info_raw)?;
+        let advisories = tcb_info.advisories(&quote_tcb_info)?;
```
nick-mobilecoin, line 35: This logic along with the common verifier below in the tests, is the main impetus for this change.
It feels like a lot to throw on every client to do and seemed like it could be better hidden/assured is done for them.

verifier/src/evidence.rs
```diff
@@ -0,0 +1,334 @@
...
+}
+
+impl<'a, Q> Accessor<Advisories> for Evidence<'a, Q> {
+    fn get(&self) -> Advisories {
+        self.advisories.clone()
+    }
+}
+
+// TODO think this should go in tcb.rs of `mc-sgx-dcap-types`
+fn tcb_info_try_from_quote<Q: AsRef<[u8]>>(quote: &Quote3<Q>) -> Result<QuoteTcbInfo> {
```
nick-mobilecoin, line 63: Like the TODO says IFF this overall PR is the path we want to go then I think we should consider adding a 
```rust
impl TryFrom<Quote3> for TcbInfo{
```
in the `mc-sgx-dcap-types`

verifier/src/evidence.rs
```diff
@@ -0,0 +1,334 @@
...
+    let signature_data = quote.signature_data();
+    let certification_data = signature_data.certification_data();
+    let CertificationData::PckCertificateChain(pem_chain) = certification_data else {
+        return Err(Error::UnsupportedQuoteCertificationData);
+    };
+    let leaf_cert = pem_chain
+        .into_iter()
+        .next()
+        .ok_or(Error::UnsupportedQuoteCertificationData)?;
+    let certificate = Certificate::from_pem(leaf_cert)?;
```
nick-mobilecoin, line 73: This feels a bit clunky to me.
Right now the `Quote3` contains certification data and that has different variants, but I'm wondering if there should be different quote variants.
This is more or less a remote attestation quote.
There are also quotes for local attestation, between enclaves on the same machine. These don't need to call out to the internet and use type 3, from https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf
> 3: Byte array that contains
concatenation of PPID encrypted using
RSA-3072-OAEP, CPUSVN, PCESVN (LE),
PCEID (LE).

verifier/src/evidence.rs
```diff
@@ -0,0 +1,334 @@
...
+        AdvisoriesVerifier, AdvisoryStatus, And, AttributesVerifier, MrSignerVerifier,
+        Quote3Verifier, TcbInfoRawVerifier, VerificationTreeDisplay, Verifier,
+    };
+    use alloc::format;
+    use core::mem;
+    use der::DateTime;
+    use mc_sgx_dcap_sys_types::{sgx_ql_ecdsa_sig_data_t, sgx_quote3_t};
+    use p256::ecdsa::VerifyingKey;
+
+    fn tcb_signing_key() -> VerifyingKey {
```
nick-mobilecoin, line 122: This and the function below are duplicates from `tcb.rs` and `quote.rs`.
I'm ok with that. I think for unit tests duplicating small pieces of logic like this are more of a boon than a bane:
- It allows the tests to be independent from each other
- It prevents creating test support function with X inputs for each test that needs a slight tweak
- Building a common infrastructure of test support methods often gets treated as second or third class, degenerating into it's own maintenance burden.

for integration tests there is a better argument for test support functions in a common place.

verifier/src/evidence.rs
```diff
@@ -0,0 +1,334 @@
...
+        )
+        .expect("Failed to decode public key");
+
+        key
+    }
+
+    type VerifierType<T> = And<
+        Quote3Verifier<T>,
+        And<MrSignerVerifier, And<AttributesVerifier, And<TcbInfoRawVerifier, AdvisoriesVerifier>>>,
+    >;
```
nick-mobilecoin, line 164: I'm not fond of this. It may be my lack of experience, but I wasn't able to get the compiler to infer the type for the below function, so punted and chose to be more explicit.

verifier/src/evidence.rs
```diff
@@ -0,0 +1,334 @@
...
+                MrSignerVerifier::new(mr_signer.into(), product_id, isv_svn.into()),
+                And::new(
+                    AttributesVerifier::new(attributes),
+                    And::new(
+                        TcbInfoRawVerifier::new(tcb_key, time),
+                        AdvisoriesVerifier::new(allowed_advisories),
+                    ),
+                ),
+            ),
+        );
```
nick-mobilecoin, line 210: Without the evidence type this would have needed to be 3 different verifiers, one for the TCB signature, one for the advisories, and another for all the quote stuff.
The manual extracting of the advisories and the 3 separate verifiers is what felt really clunky for a client.

verifier/src/evidence.rs
```diff
@@ -0,0 +1,334 @@
...
+                - [x] MRSIGNER all of the following must be true:
+                  - [x] The MRSIGNER key hash should be 0xD412_A4F0_7EF8_3892_A591_5FB2_AB58_4BE3_1E18_6E5A_4F95_AB5F_6950_FD4E_B869_4D7B
+                  - [x] The ISV product ID should be 0
+                  - [x] The ISV SVN should be at least 0
+                - [x] Both of the following must be true:
+                  - [x] The attributes should be Flags: INITTED | DEBUG | MODE64BIT Xfrm: LEGACY | AVX
+                  - [x] Both of the following must be true:
+                    - [x] The raw TCB info was verified for the provided key
+                    - [x] The allowed advisories are IDs: {"INTEL-SA-00161", "INTEL-SA-00219", "INTEL-SA-00289", "INTEL-SA-00334", "INTEL-SA-00614", "INTEL-SA-00615", "INTEL-SA-00617"} Status: OutOfDate"#;
+        assert_eq!(format!("\n{displayable}"), textwrap::dedent(expected));
```
nick-mobilecoin, line 242: This probably isn't necessary but I thought it was nice to see, especially in the error case

I'm fine to remove if desired.

verifier/src/lib.rs
```diff
@@ -163,22 +181,6 @@ pub trait Verifier<E> {
     /// In order to accommodate constant time operations this returns a
     /// [`VerificationOutput`] instead of a [`Result`].
     fn verify(&self, evidence: &E) -> VerificationOutput<Self::Value>;
```
nick-mobilecoin, line 183: The removal of the `and` and `or` methods is because they didn't actually work.
This is similar to the `VerifierType` in the evidence.rs test. I wasn't able to tweak the signature to get the compiler to infer the resultant type so one had to manually specify the type `verifier.or::<bunch of stuf here>(other)`

verifier/Cargo.toml
cbeck88: :+1:

Generally I'm against proliferation of crate features -- it frequently leads to broken builds when features are turned on in a way that isn't tested in CI, and as features proliferate you get a combinatorial explosion and you can't test them all. I usually prefer to just split things up into crates rather than have optional modules within crates, controlled by features.

In this specific case, it might be fine for the tcb feature as well to just be mandatory IMO. But that's not an issue for now

eranrund (commented):
Code-wise this looks fine to me. I like the verbose tree display thing and I like having an example output in the tests.
I don't feel like I can meaningfully comment on the quote types due to my lack of familiarity with the DCAP data types... but everything appears well documented and fairly easy to follow so I am okay with it getting merged if @cbeck88 agrees with that.

Having to manually construct that `And<...>` type is indeed suboptimal but I also couldn't figure out a way to avoid it due to wanting to use multiple traits on that thing. So returning an `impl Verifier<...>` or a `Box<dyn Verifier...>` doesn't seem to work.

How would this look in clients? Would they have to construct all this `And<>` stuff or is thrre going to be some standard verifier of some sort that we use in all our clients?

verifier/src/evidence.rs
nick-mobilecoin: I think having separate types would allow us to leverage the type system for behavior customization

The following is an oversimplification from https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf
``` 
 ┌───────────────────────────────────┐
 │                                   │
 │          Quote Header             │
 │                                   │
 ├───────────────────────────────────┤
 │                                   │
 │       Quoting Enclave Report      │
 │                                   │
 ├───────────────────────────────────┤
 │                                   │
 │       Certification Data          │
 │(Varies based on quote generation) │
 │                                   │
 └───────────────────────────────────┘
```
Right now the client gets a quote and then needs to get and try to process the certification data.

If we had a RemoteAttestationQuote, we could ensure that the Certification data was correct and thus remove the need for the clients to handle possible errors, they would instead fail to get a quote. Also the clients would be able to call `leaf_cert()` and or `cert_chain()` without digging down to the certification data and matching on it.

In the end all this may just be polish that isn't that beneficial

nick-mobilecoin:
> How would this look in clients? Would they have to construct all this `And<>` stuff or is there going to be some standard verifier of some sort that we use in all our clients?

That is something I'm still unsure of.
I'm thinking we might provide common MRSIGNER and MRENCLAVE verifier type, that would still require clients to build in the correct order, or we take in a series of arguments.
I really wish I could figure out a way to use `Box` or `dyn` here, but I was unable to display the error tree with them.

eranrund:
> > How would this look in clients? Would they have to construct all this `And<>` stuff or is there going to be some standard verifier of some sort that we use in all our clients?
> 
> That is something I'm still unsure of. I'm thinking we might provide common MRSIGNER and MRENCLAVE verifier type, that would still require clients to build in the correct order, or we take in a series of arguments. I really wish I could figure out a way to use `Box` or `dyn` here, but I was unable to display the error tree with them.

That might be possible with another trait that inherits the Verifier and tree display traits, and has a blanket implementation for everything that implements these two. Kinda hacky...

verifier/src/evidence.rs
eranrund: Okay, that makes sense. Maybe a RemoteAttestationQuote that has a TryFrom<Quote3> and extracts the certification data at that point would be nice, but it doesn't seem like a huge deal either way.

eranrund (approved):

PR #104, opened 2023-06-15
Add logic to parse JSON formatted QE identity

nick-mobilecoin:
Add logic to parse the QE(Quoting Enclave) identity from
<https://api.portal.trustedservices.intel.com/documentation#pcs-enclave-identity-v4>

Also removed the `tcb` feature. Previously the TCB logic and it's
dependencies were behind a `tcb` feature. This was to ensure this crate
suupported no alloc. While the no alloc would be nice, it isn't needed
at this time and working around it causes more work than benefit at
this time.

PR #105, opened 2023-06-16
Add `SignedQeIdentityVerifier`

nick-mobilecoin:
Add a verifier which verifies the signature and time validity of the qe
identity json data.

Be sure and look at https://github.com/mobilecoinfoundation/attestation/pull/105#discussion_r1232422318 as it may push for a different implementation


nick-mobilecoin:
* main
  * **PR #99** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/99" target="_blank">Graphite</a> 
    * **PR #104** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/104" target="_blank">Graphite</a> 
      * **PR #105** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/105" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/105?utm_source=stack-comment).

verifier/src/quote_enclave_identity.rs
```diff
@@ -107,6 +111,46 @@ pub struct SignedQeIdentity<'a> {
     signature: Vec<u8>,
 }
 
+impl<'a> SignedQeIdentity<'a> {
```
nick-mobilecoin, line 114: This is very similar to the [TcbInfoRaw](https://github.com/mobilecoinfoundation/attestation/blob/8413e9c404f158241cd854cc381e46586f48ed30/verifier/src/tcb.rs#L234) implementation.
I'm on the fence on how much to worry about the duplication.

One thought I had was:
```rust
#[derive(Debug, Deserialize, Clone)]
pub struct SignedJson<'a> {
    #[serde(borrow, alias="tcbInfo", alias="enclaveIdentity")]
    primary_field: &'a RawValue,
    #[serde(with = "hex")]
    signature: Vec<u8>,
}

#[derive(Debug, Deserialize)]
#[serde(rename_all = "camelCase")]
pub struct SignedJsonTime {
    issue_date: String,
    next_update: String,
}
```

verifier/src/quote_enclave_identity.rs
```diff
@@ -107,6 +111,46 @@ pub struct SignedQeIdentity<'a> {
...
+    }
+
+    fn verify_time(&self, time: DateTime) -> Result<(), Error> {
+        let qe_identity = QeIdentity::try_from(self)?;
+        let issue_date = qe_identity.issue_date.parse::<DateTime>()?;
+        let next_update = qe_identity.next_update.parse::<DateTime>()?;
+        if time < issue_date {
+            Err(Error::QeIdentityNotYetValid)
+        } else if time >= next_update {
+            Err(Error::QeIdentityExpired)
```
nick-mobilecoin, line 147: I didn't think about it until writing up the SignedJson comment above, but the NotYetValid and Expired are the only errors that are TCB vs QEIdentity.  The error message is "`The QE identity signature could not be verified: {error}`" so especially if the SignedJson approach is more desirable there should be no harm in genericising these errors.

verifier/src/quote_enclave_identity.rs
```diff
@@ -116,11 +160,106 @@ impl<'a> TryFrom<&'a str> for SignedQeIdentity<'a> {
...
+        let mut error = None;
+        if let Err(e) = result {
+            error = Some(e);
+        }
+
+        VerificationOutput::new(error, is_success.into())
+    }
+}
+
+impl VerificationMessage<Option<Error>> for SignedQeIdentityVerifier {
```
nick-mobilecoin, line 210: IFF the SignedJson approach is desired, I think only these verification messages would need to be implemented for each type.

verifier/src/quote_enclave_identity.rs
eranrund: I personally wouldn't worry about the duplication, it's a very small amount of simple to follow code (assuming this only repeats for these two types, if we had a bunch more then maybe I would lean towards DRYing).

verifier/src/quote_enclave_identity.rs
```diff
@@ -116,11 +160,106 @@ impl<'a> TryFrom<&'a str> for SignedQeIdentity<'a> {
...
+    type Value = Option<Error>;
+    fn verify(&self, evidence: &E) -> VerificationOutput<Self::Value> {
+        let signed_qe_identity = evidence.get();
+        let result = signed_qe_identity.verify(&self.key, self.time);
+        let is_success = result.is_ok() as u8;
+
+        let mut error = None;
+        if let Err(e) = result {
+            error = Some(e);
+        }
```
eranrund, line 204: I think this is [`result.err()`](https://doc.rust-lang.org/std/result/enum.Result.html#method.err)

eranrund (approved):
LGTM, I don't think there's a huge win in going with `SignedJson` and the `alias` trick

verifier/src/quote_enclave_identity.rs
nick-mobilecoin: done now

PR #114, opened 2023-06-26
Add QE report body verifier

nick-mobilecoin:
Add QE report body verifier

Adding a verifier that will verify the report body of a QE(quoting
enclave) to the information provided in the QE identity.

Added translation.

remove hack

improved function

fix scope

move your body every every body

more simplification

I already said I was sorry

Update docs and what not


nick-mobilecoin:
* main
  * **PR #114** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/114" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/114?utm_source=stack-comment).

verifier/src/evidence.rs
```diff
@@ -173,13 +173,13 @@ mod test {
         let quote_key = quote_signing_key(quote);
         let tcb_key = tcb_signing_key();
         let mr_signer = [
-            0xD4, 0x12, 0xA4, 0xF0, 0x7E, 0xF8, 0x38, 0x92, 0xA5, 0x91, 0x5F, 0xB2, 0xAB, 0x58,
-            0x4B, 0xE3, 0x1E, 0x18, 0x6E, 0x5A, 0x4F, 0x95, 0xAB, 0x5F, 0x69, 0x50, 0xFD, 0x4E,
-            0xB8, 0x69, 0x4D, 0x7B,
+            0x9F, 0x06, 0xDF, 0x5C, 0xA7, 0x9A, 0x23, 0xFF, 0xDF, 0xB6, 0xCA, 0x0E, 0xC8, 0x55,
```
nick-mobilecoin, line 176: This is due to the updated hw quote. Had to update it to line up with the canned qe_identity.json file.

verifier/src/qe_identity.rs
```diff
@@ -80,20 +142,68 @@ impl<'a> TryFrom<&SignedQeIdentity<'a>> for QeIdentity {
...
-struct Tcb {
+impl TcbLevel {
+    pub fn new<'a, I, E>(tcb: Tcb, tcb_status: AdvisoryStatus, advisory_ids: I) -> Self
+    where
+        I: IntoIterator<Item = &'a E>,
+        E: ToString + 'a + ?Sized,
+    {
+        Self {
+            tcb,
+            tcb_date: UNIX_TIME_STR.to_string(),
```
nick-mobilecoin, line 163: I wonder about this `tcb_date`, and some of the members in tcb.rs.
I added all the members in the json spec, but if we're not actually using them I wonder if we should just omit them.

verifier/src/qe_identity.rs
```diff
@@ -80,20 +142,68 @@ impl<'a> TryFrom<&SignedQeIdentity<'a>> for QeIdentity {
...
+        self.tcb.isv_svn.into()
+    }
+
+    pub fn status(&self) -> AdvisoryStatus {
+        self.tcb_status
+    }
+
+    pub fn advisory_ids(&self) -> &[String] {
+        &self.advisory_ids
+    }
```
nick-mobilecoin, line 179: I considered having this return `Advisories`, then doing using the `is_superset()` method in the comparison logic. Up for suggestions here, as I think there was some concern with the `is_superset()` intuitiveness to consumers.

verifier/src/qe_identity.rs
```diff
@@ -80,20 +142,68 @@ impl<'a> TryFrom<&SignedQeIdentity<'a>> for QeIdentity {
...
+}
+
+// A default value that will not be verified to be correct.
+// It will be `AdviceStatus::OutOfDate` and have an `isv_svn` of 0.
+impl Default for TcbLevel {
```
nick-mobilecoin, line 184: This felt a bit more ergonomic to have an invalid default TcbLevel, but the verification failure messages seem a bit odd, so perhaps this should go away.

verifier/src/qe_report_body.rs
```diff
@@ -0,0 +1,548 @@
...
+    }
+}
+
+impl Display for QeIsvSvnVerifier {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        write!(
+            f,
+            "The {} should correspond to an `UpToDate` level with no advisories, from: {:?}",
+            IsvSvn::spaced_struct_name(),
+            self.tcb_levels
```
nick-mobilecoin, line 236: I'm not sure if/how much we want to invest in making a Display for tcb_levels. 
I also think this output can get a bit long on the line, but really if someone is going to display the verification result it's going to be pretty big overall.

verifier/src/qe_report_body.rs
```diff
@@ -0,0 +1,548 @@
...
+    fn no_tcb_levels_for_qe_isv_svn_fails() {
+        let isv_svn = IsvSvn::from(1);
+        let verifier = QeIsvSvnVerifier::new(&[]);
+        let verification = verifier.verify(&isv_svn);
+        assert_eq!(verification.is_failure().unwrap_u8(), 1);
+
+        let displayable = VerificationTreeDisplay::new(&verifier, verification);
+        assert_eq!(
+            displayable.to_string(),
+            r#"- [ ] The ISV SVN should correspond to an `UpToDate` level with no advisories, from: [], but the ISV SVN of 1 corresponds to the TcbLevel { tcb: Tcb { isv_svn: 0 }, tcb_date: "1970-01-01T00:00:00Z", tcb_status: OutOfDate, advisory_ids: [] }"#
```
nick-mobilecoin, line 285: This message goes back to my comment early where I provided a default, over an Option<TcbLevel> it might be better to display `(none)` or similar instead of a default that the user may be wondering where it came from.

verifier/src/qe_report_body.rs
eranrund: I think we can leave it like this and if this proves to be too long then provide a nicer Display for it

verifier/src/qe_report_body.rs
eranrund: Is this referring to the entire TcbLevel? e.g. `Option<TcbLevel>`?

verifier/src/qe_report_body.rs
```diff
@@ -0,0 +1,548 @@
...
+    {
+        Self {
+            tcb_levels: tcb_levels.into_iter().cloned().collect(),
+        }
+    }
+}
+
+impl<E: Accessor<IsvSvn>> Verifier<E> for QeIsvSvnVerifier {
+    type Value = (IsvSvn, TcbLevel);
+    fn verify(&self, evidence: &E) -> VerificationOutput<Self::Value> {
```
eranrund, line 208: Does this assume the tcb_levels list is correctly sorted?

verifier/src/qe_report_body.rs
```diff
@@ -0,0 +1,548 @@
...
+        let tcb_level = tcb_levels
+            .iter()
+            .find_map(|tcb_level| {
+                if tcb_level.isv_svn().as_ref() <= isv_svn.as_ref() {
+                    Some(tcb_level.clone())
+                } else {
+                    None
+                }
+            })
+            .unwrap_or_default();
```
eranrund, line 220: Guessing this is where an Option would be used if the Default implementation went away?

eranrund (commented):
The comments made this fairly easy for me to follow, even though I am not super familiar with this part of the attestation process. Thanks for documenting it and providing elaborate tests that contain example outputs.

Got some questions sprinkled around to make sure I understand some stuff - I think the biggest thing left is to decide whether to keep the Default stuff or switch to an `Option<TcbLevel>` as an output of a failed verification?

verifier/src/qe_identity.rs
nick-mobilecoin: I've updated to use `Option<TcbLevel>` instead, this default has now been removed.

Sorry re-reading these comments, I realize I left quite a bit of context in my head and didn't write it out.

verifier/src/qe_report_body.rs
nick-mobilecoin: Yes, sorry there should have been more context in these comments

I've since moved over to the `Option<TcbLevel>`

verifier/src/qe_report_body.rs
```diff
@@ -0,0 +1,550 @@
...
+        if (!is_success).into() {
+            let name = IsvSvn::spaced_struct_name();
+            let actual = &output.value;
+
+            write!(f, ", but the {name} of {} corresponds to ", actual.0)?;
+            if let Some(tcb_level) = &actual.1 {
+                write!(f, "{tcb_level:?}")?;
+            } else {
+                write!(f, "None")?;
+            }
```
nick-mobilecoin, line 260: I went back and forth between this and
```rust
            if let Some(tcb_level) = &actual.1 {
                write!(f, ", but the {name} of {} corresponds to {tcb_level:?}", actual.0)?;
            } else {
                write!(f, ", but the {name} of {} corresponds to None", actual.0)?;
            }
```

I chose this one for DRY by default. While the alternative implementation in this comment allows for more tailoring of the "None" case, I think it's unlikely we'll see the "None" case practice as the qe identity seems to contain all the previous qe isv svn values.

verifier/src/qe_report_body.rs
nick-mobilecoin: yes it does, the step does say `(sorted by ISVSVNs)` so I think it's should be safe and potentially more robust to sort here again.  I'll work on that

nick-mobilecoin: Add logic to ensure the TCB levels are sorted.

verifier/src/qe_identity.rs
nick-mobilecoin: Sorry I left too much context in my head.
After doing a  quick prototype, I went ahead and updated to use `Advisories` and the `AdvisoriesVerifier` here in https://github.com/mobilecoinfoundation/attestation/pull/114/commits/a80ec2b4702c058ead7fba41d062dc2dc27198b7

verifier/src/qe_report_body.rs
nick-mobilecoin: Yes that is correct.
`Option<TcbLevel>` is now used here so there is no longer a need for `default`

verifier/src/qe_identity.rs
nick-mobilecoin: I think I'll leave for now. They're already here, they may help someone if they get a verification failure. Only time will tell if they are more of a bane thane a boon.

verifier/src/qe_report_body.rs
eranrund: Nit: Is `write!(f, ", but the {name} of {} corresponds to {:?}", actual.0, actual.1)?;` not viable? It will include a `Some(..)` but not sure if that's an issue

eranrund (approved):

verifier/src/qe_report_body.rs
nick-mobilecoin: it probably isn't much of an issue. 
I was probably latched on to my previous solution too much.
Updated the code to print the debug version of the `Option<T>`

nick-mobilecoin:
@nick-mobilecoin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/114).

PR #123, opened 2023-06-29
Get `Advisories` from the `Trusted**Identity` types

nick-mobilecoin:
Provide getters for getting the `Advisories` from
`TrustedMrSignerIdentity` and `TrustedMrEnclaveIdentity`. This logic
maps the two distinct config and sw hardening advisory lists into the
consolidated `Advisories` type.

### DIFF
--- a/verifier/src/identity.rs
+++ b/verifier/src/identity.rs
@@ -7,8 +7,10 @@
 //! [Intel SGX ECDSA QuoteLibReference DCAP API](https://download.01.org/intel-sgx/sgx-dcap/1.16/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf#%5B%7B%22num%22%3A63%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C69%2C468%2C0%5D)
 //! documents the identity types, Strict Policy and Security Policy.
 
+use crate::{Advisories, AdvisoryStatus};
 use alloc::string::{String, ToString};
 use alloc::vec::Vec;
+use core::ops::Not;
 use serde::{Deserialize, Serialize};
 
 /// Trusted identity for MRENCLAVE values.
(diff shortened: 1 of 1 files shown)


nick-mobilecoin:
* main
  * **PR #123** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/123" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/123?utm_source=stack-comment).

verifier/src/identity.rs
```diff
@@ -161,3 +179,120 @@ impl From<TrustedMrSignerIdentity> for TrustedIdentity {
...
+        .iter()
+        .chain(sw_hardening_advisories.iter());
+    Advisories::new(advisories, status)
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
```
nick-mobilecoin, line 207: This testing might be a bit overkill since it's a common backend, but I had the logic inverted initially so wanted to be very sure...

verifier/src/identity.rs
```diff
@@ -161,3 +179,120 @@ impl From<TrustedMrSignerIdentity> for TrustedIdentity {
         Self::MrSigner(mr_signer)
     }
 }
+
+// Convert two separate lists of config and sw hardening advisories into [`Advisories`]
+fn mitigated_advisories_to_advisories(
+    config_advisories: &[String],
+    sw_hardening_advisories: &[String],
+) -> Advisories {
+    let config_needed = config_advisories.is_empty().not();
```
eranrund, line 188: First time I'm seeing `.not()` instead of the `!` operator, but it makes sense that it works.

eranrund (approved):

verifier/src/identity.rs
nick-mobilecoin: I don't like how it's at the end, but the `!` is easy to gloss over at times

PR #124, opened 2023-06-30
Add `TrustedIdentitiesVerifier`

nick-mobilecoin:
The `TrustedIdentitiesVerifier` will verify that one of the provided
identities matches the enclave's identity.

### DIFF
--- a/verifier/src/identity.rs
+++ b/verifier/src/identity.rs
@@ -7,10 +7,14 @@
 //! [Intel SGX ECDSA QuoteLibReference DCAP API](https://download.01.org/intel-sgx/sgx-dcap/1.16/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf#%5B%7B%22num%22%3A63%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C69%2C468%2C0%5D)
 //! documents the identity types, Strict Policy and Security Policy.
 
-use crate::{Advisories, AdvisoryStatus};
+use crate::{
+    Accessor, Advisories, AdvisoriesVerifier, AdvisoryStatus, And, MrEnclaveVerifier,
+    MrSignerVerifier, VerificationOutput, Verifier,
+};
 use alloc::string::{String, ToString};
 use alloc::vec::Vec;
 use core::ops::Not;
+use mc_sgx_core_types::{IsvProductId, IsvSvn, MrEnclave, MrSigner};
 use serde::{Deserialize, Serialize};
 
 /// Trusted identity for MRENCLAVE values.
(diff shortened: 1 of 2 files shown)


nick-mobilecoin:
* main
  * **PR #124** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/124" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/124?utm_source=stack-comment).

verifier/src/identity.rs
```diff
@@ -56,8 +60,10 @@ impl TrustedMrEnclaveIdentity {
         J: IntoIterator<Item = &'a HA>,
         HA: ToString + 'a + ?Sized,
     {
+        let mut local_mr_enclave = [0u8; 32];
+        local_mr_enclave.copy_from_slice(mr_enclave.as_ref());
         Self {
-            mr_enclave: *mr_enclave,
+            mr_enclave: local_mr_enclave,
```
nick-mobilecoin, line 66: I'm contemplating removing the serde stuff from this type. as the back and forth between the primitive and the `MrEnclave` type seems error prone.
There is some argument to having a separation between the on wire type and the type used in code logic, but like most things I'm guessing "it depends"

verifier/src/identity.rs
```diff
@@ -200,14 +235,175 @@ fn mitigated_advisories_to_advisories(
...
     use super::*;
 
+    /// An identity to use in testing the different identity verifiers.
+    struct Identity {
+        mr_enclave: MrEnclave,
+        mr_signer: MrSigner,
+        isv_product_id: IsvProductId,
+        isv_svn: IsvSvn,
+        advisories: Advisories,
+    }
```
nick-mobilecoin, line 360: (first part of the comment is about the overall PR not these lines of code)

This is PR missing the `Display` implementation. 
The PR already got a bit big as it is so stopped where it's at.

I think the Display implementation  of `TrustedIdentiesVerifier` will be one of the more important ones as it's most likely that someone doesn't have new enclave Identity when they fail to attest.  I'm thinking we need to display. 

Having said that I'm thinking that in the display of a failure to verify an Identity we may want to display a type similar to this one, where the MRENCLAVE and MRSIGNER values are present.

Normally for the success case we say what was expected, but I'm thinking saying "one" of these Identities was met isn't that useful and instead we should say which one was matched.

verifier/src/identity.rs
eranrund: I'm probably missing something, but why can't the `mr_enclave` field be of type `MrEnclave` instead of a byte array?

verifier/src/identity.rs
eranrund: That all makes sense.

verifier/src/identity.rs
nick-mobilecoin: It goes back to the serde stuff
MrEnclave supports serde, but not `FromHex`

eranrund: Whats stopping it from supporting `FromHex`?

nick-mobilecoin: Nothing, just need to do it.  
I'm fine with it, if that's the path we want to take, it shouldn't be hard to do.

eranrund: Cool, I don't have a strong preference, this is all easy to back out of if needed. Just wanted to understand better whats going on. Thanks!

eranrund (approved):

PR #128, opened 2023-07-12
Add verification of `SignedTcbInfo` to `EvidenceVerifier`

nick-mobilecoin:


nick-mobilecoin:
* main
  * **PR #127** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/127" target="_blank">Graphite</a> 
    * **PR #128** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/128" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/128?utm_source=stack-comment).

verifier/src/evidence.rs
```diff
@@ -199,47 +203,133 @@ where
...
+    ) {
+        let result = self
+            .certificate_verifier
+            .verify_certificate_chain(chain, crls, self.time);
+        let is_success = result.is_ok() as u8;
+
+        // Using the default key will result in the user seeing "Error verifying the signature" for
+        // the signed data. So we try to get the key from the certificate chain, even if the
+        // verification failed. This handles the most likely failure case of an expired
+        // certificate, whose key is still the key that signed the data of interest.
```
nick-mobilecoin, line 226: Let me know if this needs more explanation. Or if there are concerns with this approach

verifier/src/evidence.rs
```diff
@@ -165,14 +168,15 @@ quote_application_report_body_field_accessor! {
...
-/// - verifying the TCB info (TODO)
+/// - verifying the TCB info
 /// - verifying the [`TrustedIdentity`] of the application enclave (TODO)
 /// - verifying the signature of the Quote (TODO)
 #[derive(Debug)]
 pub struct EvidenceVerifier<'a, C> {
     certificate_verifier: &'a C,
     _trusted_identities: Vec<TrustedIdentity>,
     time: DateTime,
+    tcb_verifier: Cell<Option<SignedTcbInfoVerifier>>,
```
nick-mobilecoin, line 179: Not fond of the interior mutability with Cell here, but we can't build the `SignedTcbInfoVerifier` until we have the `Evidence`, since the key is in the evidence.

Perhaps this verifier should be different and it should take the evidence in the constructor as well as the verify method :shrug:

verifier/src/evidence.rs
```diff
@@ -635,7 +729,71 @@ mod test {
...
+
+        let displayable = VerificationTreeDisplay::new(&verifier, verification);
+        // Note that the TCB and QE identity happen to use the same certificate, so the QE identity
+        // will also fail.
+        let expected = r#"
+            - [ ] all of the following must be true:
+              - [ ] The TCB issuer chain could not be verified: X509 certificate has been revoked
+              - [ ] The QE identity issuer chain could not be verified: X509 certificate has been revoked
+              - [x] The Quote issuer chain was verified.
+              - [x] The TCB info was verified for the provided key"#;
```
nick-mobilecoin, line 766: This test case exposed that returning the verifying key in the certificate chain verifier would result in
```
- [ ] all of the following must be true:
  - [ ] The TCB issuer chain could not be verified: X509 certificate has been revoked
  - [ ] The QE identity issuer chain could not be verified: X509 certificate has been revoked
  - [x] The Quote issuer chain was verified.
  - [ ] The TCB info could not be verified: Error verifying the signature
```

verifier/src/evidence.rs
nick-mobilecoin: I added a comment with links back to the docs for certificate chain order assumption.

I think IFF this assumed invariant is broken we need to we can put a normalizer on `Collateral` that ensures this order.

verifier/src/evidence.rs
```diff
@@ -199,47 +203,133 @@ where
...
+                        Some(CertificateChainVerifierError::GeneralCertificateError),
+                        is_success.into(),
+                    ),
+                )
+            }
+        }
     }
 }
 
+fn key_from_certificate(cert: &Certificate) -> Result<VerifyingKey, ()> {
```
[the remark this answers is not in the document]
nick-mobilecoin: Converted to an option. As mentioned here it cleaned up the usage in the calling function.

verifier/src/evidence.rs
nick-mobilecoin: Good idea!
I think I got too stuck on how it was done for other verifiers. Added the `SignedTcbInfoVerifier` to the `EvidenceValue`

verifier/src/tcb.rs
```diff
@@ -270,7 +270,7 @@ impl TryFrom<&str> for SignedTcbInfo {
 }
 
 /// Verifier for ensuring a TCB info was signed with the provided key
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq, Copy)]
```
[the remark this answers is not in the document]
nick-mobilecoin: removed now

verifier/src/evidence.rs
```diff
@@ -199,47 +203,133 @@ where
...
+        let result = self
+            .certificate_verifier
+            .verify_certificate_chain(chain, crls, self.time);
+        let is_success = result.is_ok() as u8;
+
+        // Using the default key will result in the user seeing "Error verifying the signature" for
+        // the signed data. So we try to get the key from the certificate chain, even if the
+        // verification failed. This handles the most likely failure case of an expired
+        // certificate, whose key is still the key that signed the data of interest.
+        let key = match chain.first() {
```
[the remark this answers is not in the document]
nick-mobilecoin: oh yeah that is much nicer! Especially once `key_from_certificate` has been updated to an `Option`. 

Updated.

> Unclear if that is actually more readable...

I've been trying to get more familiar with the functional chaining style, it can get a bit dense at times but I think for this case it should be easier to grok once one becomes more accustomed to the `a -> b -> c` of `foo.a().b().c()`

eranrund (approved):
LGTM

PR #129, opened 2023-07-12
Add QE identity verification to EvidenceVerifier

nick-mobilecoin:
The `EvidenceVerifier` now verifies the signed QE identity data as well
as the QE report body's identity


nick-mobilecoin:
* main
  * **PR #129** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/129" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/129?utm_source=stack-comment).

verifier/data/tests/README.md
```diff
@@ -28,10 +28,10 @@
   openssl crl -in verifier/data/tests/processor_crl.pem -out verifier/data/tests/processor_crl.der -outform DEr
   ```
 
-* `fmspc_00906ED50000_2023_05_10.json` - JSON file containing the result of a
+* `fmspc_00906ED50000_2023_07_12.json` - JSON file containing the result of a
   TCB request from
   <https://api.trustedservices.intel.com/sgx/certification/v4/tcb?fmspc=00906ED50000>.
-  This was captured on 2023-05-10.
+  This was captured on 2023-07-12.
```
nick-mobilecoin, line 34: Had to update this and the qe identity file to get them in the same time window for the verification

verifier/src/evidence.rs
```diff
@@ -317,23 +324,33 @@ impl<'a, C: CertificateChainVerifier, E: Accessor<Evidence<Vec<u8>>>> Verifier<E
...
+            qe_report_body: (qe_report_body_verifier, qe_report_body_verification),
         };
 
         let is_success = evidence_value.tcb_signing_key.is_success()
             & evidence_value.qe_identity_signing_key.is_success()
             & evidence_value.quote_signing_key.is_success()
-            & evidence_value.tcb_info.1.is_success();
+            & evidence_value.tcb_info.1.is_success()
+            & evidence_value.qe_identity.1.is_success()
+            & evidence_value.qe_report_body.1.is_success();
```
nick-mobilecoin, line 353: There is no test for a failing qe_report_body here.
I'm torn whether it's worth the effort. 
The qe identity file is signed, and so is the quote. 
To modify either of these to work with the current `EvidenceVerifier` we would need to create our own certificate chains and what not.

It may be that the chaining of all the items needs to be rethought. I.e. If this was `And::new(verifier1, And::new(verifier2, ...` then it would be assured that the `is_success()` would be accounted for.

verifier/src/evidence.rs
```diff
@@ -317,23 +324,33 @@ impl<'a, C: CertificateChainVerifier, E: Accessor<Evidence<Vec<u8>>>> Verifier<E
...
         let (_, quote_chain_verification) = self.verify_quote_signing_chain(quote, collateral);
 
         let tcb_info_verifier = SignedTcbInfoVerifier::new(tcb_key, self.time);
         let tcb_info_verification = tcb_info_verifier.verify(&evidence);
 
+        let qe_identity_verifier = SignedQeIdentityVerifier::new(qe_key, self.time);
+        let qe_identity_verification = qe_identity_verifier.verify(&evidence);
+
+        let qe_report_body_verifier = QeReportBodyVerifier::new(evidence.qe_identity.clone());
+        let qe_report_body_verification = qe_report_body_verifier.verify(&evidence);
```
nick-mobilecoin, line 337: The verifier could have been a member in `EvidenceVerifier`, but I kept it here for consistency with the other verifiers being created here.

verifier/src/evidence.rs
eranrund: I think it's acceptable to make the conscious decision to skip testing that due to the complexity of producing a broken but properly signed report body.
Another solution, which I am not sure if worth the effort, is to do something similar to what is happening [here](https://github.com/mobilecoinfoundation/mobilecoin/blob/master/transaction/builder/src/transaction_builder.rs#L818)

The idea being that you:
1. Create some trait such as `QeReportBodyVerifierTrait` that contains the `new()` method and inherits from `Verifier`
2. Implement that for `QeReportBodyVerifier` (the equivalent of `DefaultTxOutputsOrdering` in that transaction builder example)
3. Create a `fn verify_internal(&self, evidence: &E)` method inside `EvidenceVerifier` that allows passing a generic `QeReportBodyVerifierTrait` and then use that inside the implementaion
4. `fn verify()` calls `verify_internal` with `QeReportBodyVerifier`, so for all normal users the behavior and usage remain identical
5. Tests can pass their own implementation that can return whatever you want

Doesn't feel worth the effort right now...

eranrund (approved):

PR #130, opened 2023-07-12
Add `Cargo.lock`

nick-mobilecoin:
While the guidance from
<https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries>
says not to commit lock files for libraries, we've found that it is more
beneficial to have repeatable development builds. Not having a lock file
means that developers may be building with different versions of
packages. It can also lead to CI pulling in the newest changes of a
package that happens to violate major version changes.


nick-mobilecoin:
* main
  * **PR #130** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/130" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/130?utm_source=stack-comment).

eranrund (approved):

PR #131, opened 2023-07-13
Add `TrustedIdentitiesVerifier` to `EvidenceVerifier`

nick-mobilecoin:
The `EvidenceVerifier` now takes into account the provided
`trusted_identities` when verifying the provided `Evidence`.

### DIFF
--- a/verifier/src/evidence.rs
+++ b/verifier/src/evidence.rs
@@ -2,12 +2,14 @@
 
 //! The full set of evidence needed for attesting a quote
 
+use crate::identity::TrustedIdentityValue;
 use crate::qe_report_body::QeReportBodyValue;
 use crate::{
     choice_to_status_message, Accessor, Advisories, CertificateChainVerifier,
     CertificateChainVerifierError, Error, QeIdentity, QeReportBody, QeReportBodyVerifier,
     SignedQeIdentity, SignedQeIdentityVerifier, SignedTcbInfo, SignedTcbInfoVerifier, TcbInfo,
-    TrustedIdentity, VerificationMessage, VerificationOutput, Verifier, MESSAGE_INDENT,
+    TrustedIdentitiesVerifier, TrustedIdentity, VerificationMessage, VerificationOutput, Verifier,
+    MESSAGE_INDENT,
 };
 use alloc::vec::Vec;
 use core::fmt::Formatter;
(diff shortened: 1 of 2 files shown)


nick-mobilecoin:
* main
  * **PR #131** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/131" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/131?utm_source=stack-comment).

verifier/src/evidence.rs
```diff
@@ -820,12 +853,44 @@ mod test {
...
+                - [x] The allowed advisories are IDs: {"INTEL-SA-00334", "INTEL-SA-00615"} Status: SWHardeningNeeded"#;
+        assert_eq!(format!("\n{displayable}"), textwrap::dedent(expected));
+    }
+
+    #[test]
+    fn evidence_verifier_fails_for_no_identities() {
+        let time = valid_test_time();
+        let identities = [] as [TrustedIdentity; 0];
+        let certificate_verifier = TestDoubleChainVerifier::default();
+        let verifier = EvidenceVerifier::new(&certificate_verifier, identities, time);
```
nick-mobilecoin, line 893: This fails to verify for no identites, which addresses https://github.com/mobilecoinfoundation/mobilecoin/issues/3374.

I'm wondering if perhaps `EvidenceVerifier::new()` should be fallible and bark if the user provided an empty iterator of identities, since we know that will always fail.

The downside to that is it pushes more error handling on the user when creating the `EvidenceVerifier`, while there will undoubtedly need to be error handling for verification failing.

verifier/src/identity.rs
```diff
@@ -713,6 +717,28 @@ mod test {
         assert_eq!(format!("\n{displayable}"), textwrap::dedent(expected));
     }
 
+    #[test]
+    fn identities_verifier_no_identities() {
```
nick-mobilecoin, line 721: :facepalm: the "empty" test case is normally one of the test cases to always have for a sequence of items

eranrund: better late than never 😄

verifier/src/evidence.rs
dolanbernard: I agree and think this is the way to go

dolanbernard (approved):

PR #132, opened 2023-07-13
Add `Quote3Verifier` into the `EvidenceVerifier`

nick-mobilecoin:
The `EvidenceVerifier` will now verify the quote signature using a
`Quote3Verifier`.

### DIFF
--- a/verifier/src/quote.rs
+++ b/verifier/src/quote.rs
@@ -1,6 +1,6 @@
 // Copyright (c) 2023 The MobileCoin Foundation
 
-use crate::{Accessor, VerificationMessage, VerificationOutput, Verifier};
+use crate::{Accessor, Error, VerificationMessage, VerificationOutput, Verifier};
 use core::fmt::Formatter;
 use mc_sgx_dcap_types::Quote3;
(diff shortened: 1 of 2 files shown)


verifier/src/quote.rs
```diff
@@ -28,9 +28,10 @@ impl<T> Quote3Verifier<T> {
 impl<T: AsRef<[u8]>, E: Accessor<Quote3<T>>> Verifier<E> for Quote3Verifier<T> {
     type Value = ();
     fn verify(&self, evidence: &E) -> VerificationOutput<Self::Value> {
-        let quote = evidence.get();
-        let is_success = quote.verify(&self.key).is_ok() as u8;
-
+        let is_success = self
+            .key
+            .map(|key| evidence.get().verify(&key).is_ok() as u8)
+            .unwrap_or(0);
```
nick-mobilecoin, line 34: I'm ok if people feel this is too dense and should be broken out more.

eranrund: I'm fine with it, its pretty easy to follow and we've done similar stuff all over.

eranrund (approved):

PR #135, opened 2023-07-13
Add `CertificateChainVerifier` implementation

nick-mobilecoin:
A `CertificateChainVerifier` implementation is now available using
mbedtls. This implementation is behind the `x509` feature flag.

### DIFF
--- a/verifier/src/x509.rs
+++ b/verifier/src/x509.rs
@@ -1,18 +1,22 @@
 // Copyright (c) 2023 The MobileCoin Foundation
 
 extern crate alloc;
-use alloc::string::{String, ToString};
+use alloc::string::String;
 use alloc::vec;
+use alloc::vec::Vec;
 use core::fmt::{Debug, Formatter};
+use der::{DateTime, Encode};
 
 use mbedtls::{
     alloc::List as MbedtlsList,
     hash::Type as HashType,
     pk::{EcGroupId, Type as PkType},
-    x509::{Certificate, Crl, KeyUsage, Profile},
+    x509::{Certificate, Crl, Profile},
 };
 
-use x509_cert::{der::Decode, Certificate as X509Certificate};
+use crate::{CertificateChainVerifier, CertificateChainVerifierError};
+use x509_cert::crl::CertificateList;
+use x509_cert::Certificate as X509Certificate;
 
 pub type Result<T> = core::result::Result<T, Error>;
 
@@ -21,11 +25,60 @@ pub type Result<T> = core::result::Result<T, Error>;
 pub enum Error {
     /// An error occurred working with MbedTls: {0}
     MbedTls(mbedtls::Error),
+    /// An error occurred encoding to DER: {0}
+    Der(der::Error),
 }
 
 impl From<mbedtls::Error> for Error {
     fn from(src: mbedtls::Error) -> Self {
         Error::MbedTls(src)
+    }
+}
+
+impl From<der::Error> for Error {
+    fn from(src: der::Error) -> Self {
+        Error::Der(src)
+    }
+}
+
+impl From<Error> for CertificateChainVerifierError {
+    fn from(error: Error) -> Self {
+        match error {
+            // Any error (expired, revoked, etc) in certificate chain verification comes back as
+            // `X509CertVerifyFailed`
+            Error::MbedTls(mbedtls::Error::X509CertVerifyFailed) => {
+                CertificateChainVerifierError::SignatureVerification
+            }
+            _ => CertificateChainVerifierError::GeneralCertificateError,
+        }
+    }
+}
+
+/// A certificate chain verifier that uses MbedTls as the backend
+#[derive(Debug)]
+pub struct X509CertificateChainVerifier {
+    trust_anchor: TrustAnchor,
+}
+
+impl X509CertificateChainVerifier {
+    /// Create a new instance
+    pub fn new(trust_anchor: TrustAnchor) -> Self {
+        Self { trust_anchor }
+    }
+}
+
+impl CertificateChainVerifier for X509CertificateChainVerifier {
+    // Note: `_time` is ignored because MbedTls will call out to a system timer.
+    fn verify_certificate_chain<'a, 'b>(
+        &self,
+        certificate_chain: impl IntoIterator<Item = &'a X509Certificate>,
+        crls: impl IntoIterator<Item = &'b CertificateList>,
+        _time: DateTime,
+    ) -> core::result::Result<(), CertificateChainVerifierError> {
+        let unverified = UnverifiedCertChain::try_from_certificates(certificate_chain)
+            .map_err(|_| CertificateChainVerifierError::GeneralCertificateError)?;
+        let crls = CertificateRevocationList::try_from_crls(crls)?;
+        Ok(unverified.verify(&self.trust_anchor, crls)?)
     }
 }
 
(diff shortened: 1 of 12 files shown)


nick-mobilecoin:
* main
  * **PR #131** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/131" target="_blank">Graphite</a> 
    * **PR #132** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/132" target="_blank">Graphite</a> 
      * **PR #135** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/135" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/135?utm_source=stack-comment).

nick-mobilecoin:
* main
  * **PR #131** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/131" target="_blank">Graphite</a> 
    * **PR #132** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/132" target="_blank">Graphite</a> 
      * **PR #135** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/135" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/135?utm_source=stack-comment).

verifier/src/x509.rs
```diff
@@ -29,6 +35,53 @@ impl From<mbedtls::Error> for Error {
...
+                CertificateChainVerifierError::SignatureVerification
+            }
+            _ => CertificateChainVerifierError::GeneralCertificateError,
+        }
+    }
+}
+
+/// A certificate chain verifier that uses MbedTls as the backend
+#[derive(Debug)]
+pub struct X509CertificateChainVerifier {
```
nick-mobilecoin, line 59: not fond of this name, seems too wordy and possibly ambiguous with the `x509_cert` crate. Maybe mbedtls should be in the name, but that feels like it's exposing the backend that is meant to be hidden, other than mbedtls seems to be a PIA for the different SDK platforms.

verifier/src/x509.rs
```diff
@@ -1,18 +1,22 @@
 // Copyright (c) 2023 The MobileCoin Foundation
 
 extern crate alloc;
-use alloc::string::{String, ToString};
+use alloc::string::String;
```
eranrund, line 4: Nit that's not directly related to this PR but I'm surprised imports are not being grouped here, I thought we standardized on grouping them (which might require a tweak to `rustfmt.toml`).

verifier/src/x509.rs
eranrund: With the risk of bikeshedding, I think I would've chosen `MbedtlsCertificateChainVerifier` (named a feature that brings in `mbedtls` in the crate), and the file probably `mbedtls.rs` or something like that. The thinking behind that the trait `CertificateChainVerifier` already implies these are X509 certs, given that one of the arguments of its only method accepts a list of 509 certificates. So it doesn't seem like there's a world in which `CertificateChainVerifier` verifies any other kind of certificate, so the implementation is really just selecting which backend one wants to use. And as such I would've named the backend implementation that uses `mbedtls` `MbedtlsCertificateChainVerifier`.

verifier/src/x509.rs
```diff
@@ -29,6 +35,53 @@ impl From<mbedtls::Error> for Error {
...
+
+impl X509CertificateChainVerifier {
+    /// Create a new instance
+    pub fn new(trust_anchor: TrustAnchor) -> Self {
+        Self { trust_anchor }
+    }
+}
+
+impl CertificateChainVerifier for X509CertificateChainVerifier {
+    // Note: `_time` is ignored because MbedTls will call out to a system timer.
```
eranrund, line 71: Boo, not an issue for this PR to address, but that's a bummer since I think this means we can't run this code inside an enclave. I am not sure that's something we would've needed, I don't have the full picture on what enclave-to-enclave attestation looks like.

verifier/src/x509.rs
```diff
@@ -333,35 +311,56 @@ mod test {
...
-        let crl = CertificateRevocationList::try_from_pem([ROOT_CRL, PROCESSOR_CRL])
-            .expect("failed to parse CRL");
-        assert!(cert_chain.verify(&trust_anchor, crl).is_ok());
+        let crls = [ROOT_CRL, PROCESSOR_CRL]
+            .iter()
+            .map(|crl| CertificateList::from_der(crl).expect("failed to parse CRL"))
+            .collect::<Vec<_>>();
+        let verifier = X509CertificateChainVerifier::new(trust_anchor);
+        assert!(verifier
+            .verify_certificate_chain(chain.iter(), crls.iter(), time)
```
eranrund, line 327: Does the time here actually matters?

verifier/src/x509.rs
nick-mobilecoin: It's not stable :(, https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#imports_granularity
the mobilecoin repo uses nightly while we had kept this and the SGX repo on stable, with a ci build for nightly.

verifier/src/x509.rs
nick-mobilecoin: I will update to that, I wasn't really fond of the x509 naming, and I agree "certificate chain" => x509

verifier/src/x509.rs
nick-mobilecoin: nope, i missed this when I did some refactoring cleanups to use a constant `DateTime` value.  Will fix to jive with the other test cases.

verifier/src/x509.rs
nick-mobilecoin: > since I think this means we can't run this code inside an enclave.

The time in enclaves has to come from untrusted, so it's not trusted anyway :(.
However mbedtls is currently used in the enclaves, it doesn't use time, maybe I need to update the comment:
- https://github.com/mobilecoinfoundation/rust-mbedtls/blob/6d8fe323a3292f87a6bce4b35963d47139a583f9/mbedtls-sys/vendor/library/x509.c#L984 When `MBEDTLS_HAVE_TIME_DATE` is not defined.
- The default value of `MBEDTLS_HAVE_TIME_DATE`, https://github.com/mobilecoinfoundation/rust-mbedtls/blob/6d8fe323a3292f87a6bce4b35963d47139a583f9/mbedtls-sys/build/config.rs#L74
- An option to enable time/date, https://github.com/mobilecoinfoundation/rust-mbedtls/blob/6d8fe323a3292f87a6bce4b35963d47139a583f9/mbedtls-sys/build/config.rs#L420
  - It doesn't appear to be used for `"time"`, https://github.com/mobilecoinfoundation/rust-mbedtls/blob/6d8fe323a3292f87a6bce4b35963d47139a583f9/mbedtls-sys/build/features.rs#L22

verifier/src/x509.rs
nick-mobilecoin: updated now

verifier/src/x509.rs
nick-mobilecoin: updated now

verifier/src/x509.rs
nick-mobilecoin: I fixed all of them, for now...
We may need to revisit if we want to start using nightly for these repos.

verifier/src/x509.rs
nick-mobilecoin: Updated the comment to say that time is ignored all the way down for mbedtls:(.

eranrund (approved):

verifier/src/x509.rs
eranrund: Ha, shame on them. Seems like such a simple thing to stabilize 🤷‍♂️

verifier/src/x509.rs
eranrund: Makes sense, thanks for the detailed walk-through for the mbedtls implementation. Comment looks good.

PR #140, opened 2023-07-18
Advisory traits

dolanbernard:
Adding some required traits for MobileCoin integration

### Motivation

These changes are needed in order to start integrating the new attestation types into the MobileCoin repo

### DIFF
--- a/verifier/src/advisories.rs
+++ b/verifier/src/advisories.rs
@@ -19,7 +19,7 @@ use serde::{Deserialize, Serialize};
 /// use mc_attestation_verifier::AdvisoryStatus;
 /// assert!(AdvisoryStatus::UpToDate > AdvisoryStatus::SWHardeningNeeded);
 /// ```
-#[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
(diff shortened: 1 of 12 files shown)


verifier/src/advisories.rs
```diff
@@ -40,11 +40,12 @@ pub enum AdvisoryStatus {
     /// enclaves may be needed.
     SWHardeningNeeded,
     /// TCB level of the SGX platform is up-to-date.
+    #[default]
     UpToDate,
 }
 
 /// The advisories pertaining to a TCB(Trusted Computing Base).
-#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+#[derive(Debug, Default, Clone, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
```
nick-mobilecoin, line 48: Do you know where the `Ord` and `PartialOrd` are needed at?
This allows someone to do `advisories_1 < advisories_2` which for verification purposes would be good to avoid.

dolanbernard: They are needed for MrEnclaveVerifier and MrSignerVerifier along with the Kind enum in the mc repo. But it appears that those structs don't need the traits, so I'll remove it all the way down

nick-mobilecoin (approved):

PR #162, opened 2023-08-16
Export `choice_status_message`, `VerificationMessage` and `MESSAGE_INDENT`

nick-mobilecoin:
Previously the `choice_to_status_message`, `VerificationMessage` and the `MESSAGE_INDENT` were private to the crate.
This prevented consumers from being able to use these in their own `Verifier` implementations.
These items are now publicly available for use.

### DIFF
--- a/verifier/src/lib.rs
+++ b/verifier/src/lib.rs
@@ -51,8 +51,8 @@ use core::{
 };
 use subtle::Choice;
 
-/// Number of spaces to indent nested messages.
-const MESSAGE_INDENT: usize = 2;
+/// Number of spaces to indent nested [`VerificationMessage`]s.
+pub const MESSAGE_INDENT: usize = 2;
 
 /// Success checkbox indicator
 const SUCCESS_MESSAGE_INDICATOR: &str = "- [x]";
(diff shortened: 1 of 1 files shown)


verifier/src/lib.rs
```diff
@@ -60,7 +60,29 @@ const SUCCESS_MESSAGE_INDICATOR: &str = "- [x]";
...
+///     }
+/// }
+///
+/// let fail = Foo(0.into());
+/// assert_eq!(fail.to_string(), "- [ ] some message here");
+///
+/// let success = Foo(1.into());
+/// assert_eq!(success.to_string(), "- [x] some message here");
+/// ```
+pub fn choice_to_status_message(choice: Choice) -> &'static str {
```
nick-mobilecoin, line 85: Thoughts on if this should be a Trait on `Choice`? and perhaps `status_indicator`

I'm also fine leaving as is.

eranrund: Either way is fine.

eranrund (approved):

verifier/src/lib.rs
nick-mobilecoin: leaving for now. 
I need to better understand when to use traits. I think they make sense when more than one thing will need to implement the behavior, but for single extensions, functions may be better, not sure yet.

PR #169, opened 2023-09-08
Update sgx crates to 0.7.5 version

nick-mobilecoin:

PR #170, opened 2023-09-08
Add `Serialize` and `Deserialize` to `Error`

nick-mobilecoin:
Any error that is meant out be propagated out of an enclave needs to
be able to be serialized and deserialized from the trusted side into the
untrusted side.
    
`Clone` was also added as the changes needed for `Serialize` and
`Deserialize` made it possible for the error to be cloneable.

### DIFF
--- a/verifier/src/error.rs
+++ b/verifier/src/error.rs
@@ -60,3 +60,26 @@ impl From<TcbError> for Error {
         Error::Quote3TcbInfo(e)
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use assert_matches::assert_matches;
+
+    #[test]
+    fn serde_error_to_string() {
+        let bad_json = "not json";
+        let e = serde_json::from_str::<serde_json::Value>(bad_json).unwrap_err();
+        let serde_error_message = e.to_string();
+        let err = Error::from(e);
+        assert_matches!(err, Error::Serde(message) if message.contains(&serde_error_message));
+    }
+
+    #[test]
+    fn der_error_to_string() {
+        let e = der::Error::incomplete(1u8.into());
+        let der_error_message = e.to_string();
+        let err = Error::from(e);
+        assert_matches!(err, Error::Der(message) if message.contains(&der_error_message));
+    }
+}
--- a/verifier/src/error.rs
+++ b/verifier/src/error.rs
@@ -2,15 +2,17 @@
 
 //! Errors that can occur during verification
 
+use alloc::string::{String, ToString};
 use mc_sgx_dcap_types::TcbError;
+use serde::{Deserialize, Serialize};
 
 /// Error working with quote evidence
-#[derive(displaydoc::Display, Debug)]
+#[derive(displaydoc::Display, Debug, Clone, Serialize, Deserialize)]
 pub enum Error {
     /// Error converting from DER {0}
-    Der(der::Error),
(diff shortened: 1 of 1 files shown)


nick-mobilecoin:
* main
  * **PR #169** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/169" target="_blank">Graphite</a> 
    * **PR #170** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/170" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/170?utm_source=stack-comment).

Cargo.toml
```diff
@@ -22,6 +22,11 @@ lto = "thin"
 mbedtls = { git = "https://github.com/mobilecoinfoundation/rust-mbedtls.git", rev = "6d8fe323a3292f87a6bce4b35963d47139a583f9" }
 mbedtls-sys-auto = { git = "https://github.com/mobilecoinfoundation/rust-mbedtls.git", rev = "6d8fe323a3292f87a6bce4b35963d47139a583f9" }
 
+mc-sgx-dcap-types = { git = "https://github.com/mobilecoinfoundation/sgx.git", rev = "b86126432d5e5a7705bb7be9d3358963910f2ccc" }
```
nick-mobilecoin, line 25: These will need to go away prior to this merging, waiting on https://github.com/mobilecoinfoundation/sgx/pull/417

verifier/src/error.rs
```diff
@@ -49,7 +50,7 @@ impl From<der::Error> for Error {
 
 impl From<serde_json::Error> for Error {
     fn from(e: serde_json::Error) -> Self {
-        Error::Serde(e)
+        Error::Serde(e.to_string())
```
nick-mobilecoin, line 53: I contemplated adding a test like
```rust
#[cfg(test)]
mod test {
    use super::*;

    #[test]
    fn serde_error_to_string() {
        let bad_json = "not json";
        let e = serde_json::from_str::<serde_json::Value>(bad_json).unwrap_err();
        let err = Error::from(e);
        assert_eq!(err.to_string(), "Error parsing TCB(Trusted Computing Base) json info: expected ident at line 1 column 2");
    }
}
```

but it feels too fragile to changes in serde's error string formatting, and that's really what a test for this should be testing that the serde error string is propagated properly.
I'm not opposed to adding it if it's desired.

dolanbernard: What about something like
```Rust
err.to_string().contains(e.to_string())
```

dolanbernard (approved):

verifier/src/error.rs
nick-mobilecoin: Good suggestion.
Added a test case similar to that suggestion in, https://github.com/mobilecoinfoundation/attestation/pull/170/commits/fcde25983083e9c37d21a73bb423f123eb7b6bd9

Cargo.toml
nick-mobilecoin: removed now

nick-mobilecoin:
### Merge Activity

* **Sep 21, 5:43 PM**: @nick-mobilecoin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/attestation/170).

PR #177, opened 2023-09-20
Time is now Optional

dolanbernard:
* Make time Optional for cert validation.

### Motivation

Not all mc code has access to time and isn't able to verify issue date/expiration. A None value for time can be used to signal that time verification cannot occur.


verifier/src/qe_identity.rs
```diff
@@ -646,6 +677,7 @@ mod test {
         let time = "2023-07-12T20:48:25Z"
             .parse::<DateTime>()
             .expect("Failed to parse time");
+        let time = Some(time);
```
dolanbernard, line 680: I have a lot of these in the test code. Not sure how much it matters, but it might look confusing reassigning time immediately. I like this better than the alternative because the linter forces you to add a weird comma to the end of `Some` blocks.

verifier/src/evidence.rs
```diff
@@ -507,10 +512,12 @@ mod test {
...
     // Valid time for the TCB_INFO_JSON and QE_IDENTITY_JSON
-    fn valid_test_time() -> DateTime {
-        "2023-07-12T20:48:25Z"
-            .parse::<DateTime>()
-            .expect("Failed to parse time")
+    fn valid_test_time() -> Option<DateTime> {
+        Some(
+            "2023-07-12T20:48:25Z"
+                .parse::<DateTime>()
+                .expect("Failed to parse time"),
```
dolanbernard, line 519: The linter enforced comma from my other comment

verifier/src/mbedtls.rs
```diff
@@ -322,7 +322,7 @@ mod test {
             .collect::<Vec<_>>();
         let verifier = MbedTlsCertificateChainVerifier::new(trust_anchor);
         assert!(verifier
-            .verify_certificate_chain(chain.iter(), crls.iter(), DateTime::INFINITY)
+            .verify_certificate_chain(chain.iter(), crls.iter(), None)
```
dolanbernard, line 325: It looks to me like we're skipping time validation here anyway, so this should be fine

verifier/src/certificate_chain.rs
```diff
@@ -33,11 +33,13 @@ pub trait CertificateChainVerifier {
     /// * `crls` - The certificate revocation lists to use when verifying the certificate chain.
     /// * `time` - The time to use when verifying the certificate chain. Due to implementation
     ///   details, some implementations may ignore this value and use the system time directly.
+    ///   A None value for time can be used in cases where the calling unable to provide time. In
+    ///   such cases, time validation must be skipped.
```
eranrund, line 37:
```suggestion
    ///   A None value for time can be used in cases where the caller is unable to provide time. In
    ///   such cases, time validation will be skipped.
```

verifier/src/evidence.rs
```diff
@@ -191,8 +191,13 @@ where
     ///   of the identities needs to match to succeed for the identity verification portion.
     /// * `time` - The time to use for verifying the evidence. In particular the TCB Info and QE
     ///   identity have expiry times that need to be verified. Note: that the `certificate_verifier`
-    ///   will also be passed this time.
-    pub fn new<I, ID>(certificate_verifier: C, trusted_identities: I, time: DateTime) -> Self
+    ///   will also be passed this time. A None value for time can be used in cases where the calling
+    ///   code is unable to provide time. In such cases, time validation must be skipped.
```
eranrund, line 195:
```suggestion
    ///   code is unable to provide time. In such cases, time validation will be skipped.
```

verifier/src/qe_identity.rs
```diff
@@ -296,13 +297,14 @@ impl SignedQeIdentityVerifier {
     ///   QE identity request from
     ///   <https://api.trustedservices.intel.com/sgx/certification/v4/qe/identity?update=standard>
     ///   When this is `None` verification will always fail.
-    /// - `time` - The current system time
+    /// - `time` - The current system time. A None value for time can be used in cases where the
+    ///   calling code is unable to provide time. In such cases, time validation must be skipped.
```
eranrund, line 301:
```suggestion
    ///   calling code is unable to provide time. In such cases, time validation will be skipped.
```

eranrund (commented):
LGTM, but going to let @nick-mobilecoin review as well since he wrote most of this.

verifier/src/mbedtls.rs
nick-mobilecoin: that's correct, Mbedtls will only verify the time if MBEDTLS_HAVE_DATE_TIME is set, https://github.com/Mbed-TLS/mbedtls/blob/5edb9427086a4cdc8e567bb8069db9d09a58e21b/library/x509.c#L1109

Our build leaves it undefined, https://github.com/mobilecoinfoundation/rust-mbedtls/blob/mc-develop/mbedtls-sys/build/config.rs#L74

Mbedtls also uses a function pointer to get the time value when it verifies the certificates

verifier/src/qe_identity.rs
```diff
@@ -136,21 +136,22 @@ impl QeIdentity {
         sgx_attributes_t { flags, xfrm }.into()
     }
 
-    fn verify(&self, time: DateTime) -> Result<(), Error> {
+    fn verify(&self, time: Option<DateTime>) -> Result<(), Error> {
```
nick-mobilecoin, line 139: Note if you do `impl Into<Option<DateTime>>` for the public functions, I would probably leave these private ones as `Option<DateTime>`

verifier/src/qe_identity.rs
```diff
@@ -136,21 +136,22 @@ impl QeIdentity {
...
-        } else {
-            Ok(self)
+        if let Some(date_time) = time {
+            if date_time < issue_date {
+                return Err(Error::QeIdentityNotYetValid);
+            } else if date_time >= next_update {
+                return Err(Error::QeIdentityExpired);
+            }
         }
+        Ok(self)
```
nick-mobilecoin, line 154: Not sure it's better, I could see arguments for concern on that final `Some(_)`
```suggestion
        match time {
            Some(date_time) if date_time < issue_date => Err(Error::QeIdentityNotYetValid),
            Some(date_time) if date_time >= next_update => Err(Error::QeIdentityExpired),
            None | Some(_) => Ok(self),
        }
```

verifier/src/qe_identity.rs
```diff
@@ -248,7 +249,7 @@ impl SignedQeIdentity {
     ///     let time = DateTime::from_system_time(SystemTime::now()).unwrap();
     ///     ```
     ///   or equivalent
-    pub fn verify(self, key: Option<&VerifyingKey>, time: DateTime) -> Result<(), Error> {
+    pub fn verify(self, key: Option<&VerifyingKey>, time: Option<DateTime>) -> Result<(), Error> {
```
nick-mobilecoin, line 252: IFF `impl Into<Option<DateTime>>` is used elsewhere I would suggest doing it here too because it's public, even though it's likely to only be used by evidence which knows what it has
```suggestion
    pub fn verify(self, key: Option<&VerifyingKey>, time: impl Into<Option<DateTime>>) -> Result<(), Error> {
```

verifier/src/qe_identity.rs
```diff
@@ -296,13 +297,14 @@ impl SignedQeIdentityVerifier {
...
-    /// - `time` - The current system time
+    /// - `time` - The current system time. A None value for time can be used in cases where the
+    ///   calling code is unable to provide time. In such cases, time validation must be skipped.
     ///   This is expected to be generated by:
     ///     ```ignore
     ///     let time = DateTime::from_system_time(SystemTime::now()).unwrap();
     ///     ```
     ///   or equivalent
-    pub fn new(key: Option<VerifyingKey>, time: DateTime) -> Self {
+    pub fn new(key: Option<VerifyingKey>, time: Option<DateTime>) -> Self {
```
nick-mobilecoin, line 307: similar here IFF impl INto is used
```suggestion
    pub fn new(key: Option<VerifyingKey>, time: impl Into<Option<DateTime>>) -> Self {
```

verifier/src/qe_identity.rs
```diff
@@ -555,7 +557,7 @@ mod test {
         let signed_qe_identity =
             SignedQeIdentity::try_from(json).expect("Failed to parse signed identity");
         let key = qe_verifying_key();
-        let time = time.parse::<DateTime>().expect("Failed to parse time");
+        let time = Some(time.parse::<DateTime>().expect("Failed to parse time"));
```
nick-mobilecoin, line 560: IFF impl into is used all these `Some()`s could go away

verifier/src/qe_identity.rs
```diff
@@ -612,12 +616,39 @@ mod test {
...
+        let signed_qe_identity =
+            SignedQeIdentity::try_from(json).expect("Failed to parse signed identity");
+        let key = qe_verifying_key();
+        let verifier = SignedQeIdentityVerifier::new(Some(key), None);
+
+        let verification = verifier.verify(&signed_qe_identity);
+
+        assert_eq!(verification.is_success().unwrap_u8(), 1);
+    }
+
```
nick-mobilecoin, line 644: This looks exactly the same as the previous test, I don't think it's providing anything.
```suggestion
```

verifier/src/qe_identity.rs
```diff
@@ -612,12 +616,39 @@ mod test {
         );
     }
 
+    #[test]
+    fn qe_identity_verifier_passes_at_next_update_ignoring_time() {
```
nick-mobilecoin, line 620: I don't5 think "at next update" matters here
```suggestion
    fn qe_identity_verifier_passes_ignoring_time() {
```

verifier/src/qe_identity.rs
nick-mobilecoin: that linter is odd, I would probably do it in the function call, but IFF the `Into<Option` path is chosen they could all go away too.
```rust
let verifier = SignedQeIdentityVerifier::new(Some(key), Some(time));
```

verifier/src/tcb.rs
```diff
@@ -672,20 +674,42 @@ mod tests {
...
+        let time = Some(time);
 
         assert_matches!(
             signed_tcb_info.verify(Some(&key), time),
             Err(Error::TcbInfoExpired)
         );
     }
 
+    #[test]
+    fn passes_before_issue_date_ignoring_time() {
```
nick-mobilecoin, line 686: don't think "before" matters since time is None
```suggestion
    fn passes_ignoring_time() {
```

verifier/src/tcb.rs
```diff
@@ -672,20 +674,42 @@ mod tests {
...
+    #[test]
+    fn passes_at_next_update_ignoring_time() {
+        let key = tcb_verifying_key();
+        let tcb_json = include_str!("../data/tests/fmspc_00906ED50000_2023_07_12.json");
+        let signed_tcb_info =
+            SignedTcbInfo::try_from(tcb_json).expect("Failed to parse signed TCB");
+
+        assert!(signed_tcb_info.verify(Some(&key), None).is_ok());
+    }
+
```
nick-mobilecoin, line 704: Think this is the same as teh test above and can go away
```suggestion
```

verifier/src/tcb.rs
```diff
@@ -106,21 +106,22 @@ impl TcbInfo {
...
+        if let Some(date_time) = time {
+            let issue_date = self.issue_date.parse::<DateTime>()?;
+            let next_update = self.next_update.parse::<DateTime>()?;
+            if date_time < issue_date {
+                return Err(Error::TcbInfoNotYetValid);
+            } else if date_time >= next_update {
+                return Err(Error::TcbInfoExpired);
+            }
         }
+        Ok(self)
```
nick-mobilecoin, line 124: IFF the qe_identity logic is updated
```suggestion
        let issue_date = self.issue_date.parse::<DateTime>()?;
        let next_update = self.next_update.parse::<DateTime>()?;
        match time {
            Some(date_time) if date_time < issue_date => Err(Error::TcbInfoNotYetValid),
            Some(date_time) if date_time >= next_update => Err(Error::TcbInfoExpired),
            None | Some(_) => Ok(self),
        }
```

nick-mobilecoin (approved):
looks good, mainly some thoughts on potential ergonomics to the API

verifier/src/qe_identity.rs
dolanbernard: Yeah, I was wondering what it would look like to do this with pattern matching. I don't know if I like this. In my opinion this is a case of just because we can use a feature it doesn't mean we should. I think the original is more readable, but I could be convinced otherwise.

verifier/src/qe_identity.rs
dolanbernard: Ah, yes, I think I got a little ahead of myself there. Indeed there is only one path to test here. I'll get rid of both duplicate tests. Good point.

verifier/src/certificate_chain.rs
```diff
@@ -33,11 +33,13 @@ pub trait CertificateChainVerifier {
...
     ///
     fn verify_certificate_chain<'a, 'b>(
         &self,
         certificate_chain: impl IntoIterator<Item = &'a Certificate>,
         crls: impl IntoIterator<Item = &'b CertificateList>,
-        time: DateTime,
+        time: Option<DateTime>,
```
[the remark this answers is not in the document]
dolanbernard: Ah, this is a very good idea and will save lots of work later

verifier/src/evidence.rs
dolanbernard: The impl Into changes fixes this dumb comma :partying_face:

nick-mobilecoin (approved):
LGTM

verifier/src/qe_identity.rs
nick-mobilecoin: sounds good, I don't have a strong preference either way, the match guards can be a little dense to read at times.

PR #222, opened 2024-04-05
Bump nightly CI runs to 10-01-2023

nick-mobilecoin:
This matches the nightly currently used by the main consumer of this
crate,
<https://github.com/mobilecoinfoundation/mobilecoin/blob/39fbbdaae314c5696f44bfb4df1c46d75ce46aeb/rust-toolchain#L2>

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,10 +10,6 @@ name: ci
 env:
   CARGO_INCREMENTAL: 0
   RUSTFLAGS: "-Dwarnings"
-  # This works around `dtolnay/rust-toolchain@master` it defaults to "sparse"
-  # when 1.68 is used, but the current nightly version (nightly-2023-01-04) is
-  # before the official support that doesn't use `-Z sparse`
-  CARGO_REGISTRIES_CRATES_IO_PROTOCOL: "git"
 
 jobs:
   # TODO: Fix automatically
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -10,10 +10,6 @@ name: ci
 env:
   CARGO_INCREMENTAL: 0
   RUSTFLAGS: "-Dwarnings"
-  # This works around `dtolnay/rust-toolchain@master` it defaults to "sparse"
```
nick-mobilecoin, line 13: per the comment this was only needed for the 2023-01-04 nightly

.github/workflows/ci.yaml
```diff
@@ -35,92 +31,6 @@ jobs:
       # FIXME: Add yamllint problem matcher
       - run: yamllint -s .
 
-  crev:
```
nick-mobilecoin, line 38: Figured while cleaning up other things it might be good to remove steps that aren't currently benefiting our CI.
While crev or something like it sounds good it's too much of a lift at this time to try and review all transitive dependencies, we're not in a place and likely not going to be in a place where this check becomes actionable. Becuase of that this check is just noise at this time.

.github/workflows/ci.yaml
```diff
@@ -259,35 +169,3 @@ jobs:
...
-      - lint
-      - deny
-      - sort
-      - clippy
-      - build
-      - test
-      - doc
-      - coverage
-    steps:
-      - name: Notify Discord on failure
```
nick-mobilecoin, line 276: We've pulled back from doing dev/eng efforts in discord, so this only creates noise that doesn't help at this time.
While it would be nice to have a slack notification I don't think it's worth the effort at this time to implement

jgreat (approved):

